PR #120, opened 2020-03-30
Extract current artifacts writing to a separate class

mike1808:
* Extract artifact serialization to ArtifactWriter
* Artifacts now only represent the tracked record and they itself do not
handle their storing process except binary artifacts: images and models
* Artifacts storing logic is moved to ArtifactWriter class
* Serializable#serialize -> Artifact#serialize now returns a Record or
RecordCollection

The work is still in progress. Next steps are:

1. Create an abstract `RecordWriter` class
2. Implement `LogRecordWrtier` using the current logic of saving artifacts from `AimRepo`.


gorarakelyan:
- Created `RecordWriter` class which handles storing artifact records
- Moved `json` and `json_logs` records storing logic from `AimRepo` to `RecordWriter`
- Added aimrecords + protobuf serialization support for `Artifact`s

PR #121, opened 2020-04-13
Add autocompletion to branch, push, remote commands

mike1808:
To enable autocompletion see [autocompletion
docs](https://click.palletsprojects.com/en/7.x/bashcomplete/#activation)


aim/cli/branch/commands.py
```diff
@@ -1,9 +1,11 @@
 import click
 
+from aim.cli.branch.utils import get_branches
+
 
 @click.group()
 @click.pass_obj
-def branch_entry_point(repo):
+def branch(repo):
```
mike1808, line 8: Those entrypoints were renamed because autocompletion was using the function names as autocompletion prompts.

aim/cli/view/commands.py
```diff
@@ -8,7 +8,7 @@
 
 @click.group()
 @click.pass_obj
-def view_entry_point(repo):
+def view(repo):
```
gorarakelyan, line 11: Maybe we also have to rename `aim board` command entry function name to `board` or is it ok?

mike1808: What do you mean? I cannot find `aim board` command.

gorarakelyan: When board functionality was added the command name was `aim view` but then we decided to name it board and i just changed the command name in configs(`VIEW_NAME = 'board'`). So I guess it should be renamed as well.

mike1808: @gorarakelyan I see what do you mean. So, right now autocomplete returns "view" because it doesn't care about name argument passed in `cli_entry_point.add_command(view_commands.view, VIEW_NAME)`. So we have two options:

1. Rename this function to `board`
1. Use our command name constants as the first argument in `@cli.group()` and `@cli.command()` decorators. 

For now. I'll just rename the function and package name. But doing that I'm just making two sources of truth: command name from decorator, command name from `add_command` option. 

I'd say that it's `click` bug that autocomplete doesn't know about `add_command` arguments, but anyway, I don't like the current approach of renaming functions just to rename cli command name.

gorarakelyan: Got it, cool.

re "two sources of truth": Yea, a little bit weird, but I guess renaming is the simplest solution for now. We can refactor this piece in future, if we need it.

PR #134, opened 2020-05-16
added text.py

jamesj-jiao:


aim/sdk/artifacts/text.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Text(Artifact, metaclass=ABCMeta):
+    cat = ('text',)
+    _step_counter = {}
+
+    def __init__(self, document_name: str, text: str):
+        self.document_name = document_name
+        self.text = text
+        self.timestamp = int(time.time())
+        self.json = {"text": self.text, "timestamp": self.timestamp}
```
gorarakelyan, line 16: Maybe we should rename `json` property to smth more specific, e.g. `content`? And move this line to`serialize` method, cause constructing actual content is a part of "serilization", right?

PR #135, opened 2020-05-16
update text.py with more specific var names

jamesj-jiao:


examples/track_text.py
```diff
@@ -0,0 +1,4 @@
+import aim
+from aim import track
+
+track(aim.text, "document_name", "text")
```
gorarakelyan, line 4: Lets track values inside a for loop to demonstrate that values are sequentially appended to the document `document_name`

aim/engine/configs.py
```diff
@@ -16,6 +16,7 @@
 AIM_PARAMS_DIR_NAME = 'hyperparameters'
 AIM_MAP_DIR_NAME = 'map'
 AIM_STATS_DIR_NAME = 'stats'
+AIM_TEXT_DIR_NAME = 'document_text'
```
gorarakelyan, line 19: Lets have directory names in plural form, cause 1 directory can store a collection of several artifacts

PR #141, opened 2020-05-24
Create PEP8 Config File and Enable Recursive Checking & Formatting

jialin-wu-02:
- Create PEP8 config files, excluding checks in lib.
- Enable recursive checking with `pycodestyle` and formatting with `autopep8`.
- Update README to include guide for formatting code.

### DIFF
--- a/aim/cli/cli.py
+++ b/aim/cli/cli.py
@@ -1,6 +1,5 @@
 import click
 from click import core
-core._verify_python3_env = lambda: None
 
 from aim.engine.aim_repo import AimRepo
 
@@ -16,6 +15,8 @@ from aim.cli.version import commands as version_commands
 from aim.cli.view import commands as view_commands
 from aim.cli.status import commands as status_commands
 from aim.cli.config import commands as config_commands
+
+core._verify_python3_env = lambda: None
 
 
 @click.group()
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,3 @@
+[pycodestyle]
+exclude=lib, ./aim/sdk/artifacts/proto, ./examples, ./tests/models
+max-line-length = 80
--- a/README.md
+++ b/README.md
@@ -32,3 +32,10 @@ pip install -r requirements.txt
 ├── examples <------------  example usages of aim SDK
 └── tests
 ```
+
+### Code Style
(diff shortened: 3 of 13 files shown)


jialin-wu-02:
Include quite a lot of file changes since I have formatted all the code again.

setup.cfg
```diff
@@ -0,0 +1,3 @@
+[pep8]
+exclude=lib
```
gorarakelyan, line 2: Lets also exclude protobuf generated files(`sdk/artifacts/proto`)

gorarakelyan: Maybe we should also exclude `examples`? Because they are mostly downloaded from tutorials or open source projects and would be better to keep them in the original style.

setup.cfg
```diff
@@ -0,0 +1,3 @@
...
+max-line-length = 100
```
gorarakelyan, line 3: Lets change it to 80 [https://www.python.org/dev/peps/pep-0008/#maximum-line-length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

setup.cfg
gorarakelyan: Maybe we should also exclude examples? Because they are mostly downloaded from tutorials or open source projects and would be better to keep them in the original style.

aim/cli/cli.py
```diff
@@ -1,22 +1,20 @@
...
+from aim.cli.reset import commands as reset_commands
+from aim.cli.commit import commands as commit_commands
+from aim.cli.branch import commands as branch_commands
+from aim.cli.push import commands as push_commands
+from aim.cli.remote import commands as remote_commands
+from aim.cli.init import commands as init_commands
+from aim.cli.auth import commands as auth_commands
+from aim.cli.configs import *
+from aim.engine.aim_repo import AimRepo
 import click
```
gorarakelyan, line 14: Dependencies are imported in this order:
- 1st block: system, os, default deps
- 2nd block: 3rd party deps (can be combined with 1st block if few packages are imported) 
- 3rd block: aim deps

Blocks are separated by blank line.

jialin-wu-02:
Ignore `./aim/sdk/artifacts/proto`, `./examples`, `./tests/models` now, since they are downloaded examples. Also slightly alter how we run `autopep8`. Now the changes seem more reasonable.

PR #142, opened 2020-05-24
[WIP] Add tensorflow 1.x integration to track wights & biases distributions via aim SDK.

sevakharutyunyan:
Following the pattern of names and logic in the `examples` package, `tf_track_distribution.py` is added. Also, corresponding changes are made in the `aim/sdk/artefacts/distribution` module. For now, we will just assume, that if the NN module is not derived from `pytorch`, then it is created by `tensorflow`. For complete tracking, layer names should be added too. 

**PR UPDATE 2**
Layer names are track via low API `tf.layers`. It works for any model, which layers are instantiated by `tf.layers`. 
 
Next steps:
[+] - Track graph operation names
[+]- Track graph trainable variables' values
[  ] - Add condition checking if model is built in tensorflow

**PR UPDATE 3**
`track(<artefact>, session)` is applied. API takes  tf.Session as input, and tracks weights and biases for both ops and layers. 

Next steps:
1. Notify users about  layer name default generation (as in `tf`), if they didn't name their variables, or didn't enscope the variables or operations.

### DIFF
--- a/aim/sdk/artifacts/distribution.py
+++ b/aim/sdk/artifacts/distribution.py
@@ -4,7 +4,7 @@ from typing import Any
 from aim.engine.utils import is_pytorch_module, get_module
 from aim.sdk.artifacts.artifact import Artifact
 from aim.sdk.artifacts.record import Record, RecordCollection
-from aim.sdk.artifacts.utils import get_pt_tensor
+from aim.sdk.artifacts.utils import get_pt_tensor, TfUtils
 
 
 class Distribution(Artifact):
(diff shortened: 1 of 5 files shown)


aim/sdk/artifacts/distribution.py
```diff
@@ -120,7 +120,25 @@ def get_layers(cls, model, parent_name=None):
                             bias_hist[0].tolist(),
                             bias_hist[1].tolist(),
                         ]
-
+        else:
```
sevakharutyunyan, line 123: here should go
```python
if is_tf_module(model):
```

aim/sdk/artifacts/distribution.py
```diff
@@ -120,7 +120,25 @@ def get_layers(cls, model, parent_name=None):
...
+            params = model
+            num_of_layers = int(len(params) / 2)
+            weights = model[num_of_layers: ]
+            biases = model[: num_of_layers]
+            for i in range(num_of_layers):
+                layer_name = f"Layer{i+1}"
```
sevakharutyunyan, line 129: Here is the `layer name` issue.

examples/tf_track_distribution.py
```diff
@@ -0,0 +1,100 @@
+import aim
+from aim import track
+
+import tensorflow as tf
+
+# Import MNIST data
+from tensorflow.examples.tutorials.mnist import input_data
```
sevakharutyunyan, line 7: I've got problems with this line. See [here](https://github.com/tensorflow/tensorflow/issues/28801)

PR #144, opened 2020-05-26
Refactor Text artifact to store sentences using protobuf and aimrecords

jamesj-jiao:
Update Changelog


gorarakelyan:
Argument `create_key=_descriptor._internal_create_key` inside `text_pb2.py` raises an incompatibility error with current libproto version. Just remove that to resolve the issue.

PR #146, opened 2020-06-02
Add keras 2.x integration to track wights & biases distributions of ML model via aim SDK.

vb99:
Edit 1: Integrated keras 2.x with aim using aim sdk by editing the get_layers method in aim/sdk/artifacts/distribution.py. I was able to access the weights and biases in the keras model and used the layers dictionary to hold the distribution.

Edit 2: Created examples/keras_track_distribution.py which runs an example MNIST Keras model and tracks the weights distribution at the end of the training.

### DIFF
--- a/examples/keras_track_distribution.py
+++ b/examples/keras_track_distribution.py
@@ -1,0 +1,56 @@
+import aim
+from aim import track
+
+import numpy as np
+import mnist
+from keras.models import Sequential
+from keras.layers import Dense
+from keras.utils import to_categorical
+
+"""
+Citation for Keras Example Model: victorzhou.com/blog/keras-neural-network-tutorial/
+"""
+
+train_images = mnist.train_images()
+train_labels = mnist.train_labels()
+test_images = mnist.test_images()
+test_labels = mnist.test_labels()
+
+# Normalize the images.
+train_images = (train_images / 255) - 0.5
+test_images = (test_images / 255) - 0.5
+
+# Flatten the images.
+train_images = train_images.reshape((-1, 784))
+test_images = test_images.reshape((-1, 784))
+
+# Build the model.
+model = Sequential([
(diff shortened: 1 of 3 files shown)


examples/keras_track_distribution.py
```diff
@@ -0,0 +1,56 @@
...
+  batch_size=32,
+)
+
+# Evaluate the model.
+model.evaluate(
+  test_images,
+  to_categorical(test_labels)
+)
+
+weights, biases = model.layers[0].get_weights()
```
gorarakelyan, line 55: To track distributions per iteration or per epoch the callback function should be passed to model.fit function which will call `track(aim.weights, model)` from inside. You can learn more about keras callbacks here: https://keras.io/api/callbacks/

vb99: Hi, I have added a callback class to track weights distributions every epoch.

examples/keras_track_distribution.py
```diff
@@ -0,0 +1,59 @@
...
+from keras.models import Sequential
+from keras.layers import Dense
+from keras.utils import to_categorical
+
+"""
+Citation for Keras Example Model: victorzhou.com/blog/keras-neural-network-tutorial/
+"""
+
+class TrackCallBack(keras.callbacks.Callback):
+  def on_epoch_end(self, epoch, logs=None):
```
gorarakelyan, line 16: - It would be better to not only track weights on epoch end, but also on iteration end and on some condition(e.g. on each 2*k iteration).
 - Also it would be good to move `TrackCallBack` into SDK and make it importable, so users can just import and use it.

vb99: Hi- Now, I track weights on epoch end and every k-th iteration end. I have also moved TrackCallBack in track.py in SDK so that it is importable.

PR #153, opened 2020-06-12
Add tensorflow 1.x integration to track weights & biases distributions via aim SDK.

bhaveshk658:
From the previous pull request, a check for whether the module was created in TensorFlow was added. In addition, the order of the names of the layers is no longer sorted alphabetically when viewing through the board to maintain the ordering of the NN.

### DIFF
--- a/aim/engine/utils.py
+++ b/aim/engine/utils.py
@@ -121,7 +121,12 @@ def is_pytorch_module(inst):
     """
     return inst_has_typename(inst, ['torch', 'Module'])
 
-
+def is_tensorflow_session(inst):
+    """
+    Check whether `inst` is isntance of tensorflow session
+    """
+    return inst_has_typename(inst, ['session'])
+    
 def is_pytorch_optim(inst):
     """
     Check whether `inst` is instance of pytorch optimizer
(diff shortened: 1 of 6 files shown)


aim/engine/utils.py
```diff
@@ -121,7 +121,12 @@ def is_pytorch_module(inst):
     """
     return inst_has_typename(inst, ['torch', 'Module'])
 
-
+def is_tensorflow_session(inst):
+    """
+    Check whether `inst` is isntance of tensorflow session
+    """
+    return inst_has_typename(inst, ['session'])
```
gorarakelyan, line 128: Is only checking session enough? Maybe we should also check if an instance has `tf`, `tensorflow` in its class type

gorarakelyan:
@bhaveshk658 Have you also tested for `neural_net_with_ops` model?

bhaveshk658:
I think it looks alright with that model as well:
Screen Shot 2020-06-12 at 9 40 43 AM

aim/engine/utils.py
bhaveshk658: Added 'tensorflow' to types.

gorarakelyan:
Good job!

PR #158, opened 2020-06-18
Add tensorflow 1.x integration with checkpoints and saving/loading models.

bhaveshk658:
The ability to track checkpoints and save models in TensorFlow 1.x has been added using the Saver class following the existing code for PyTorch and Keras. In addition, examples have been added to test saving and loading models in TensorFlow.

Loading models produces the error `tensorflow.python.framework.errors_impl.DataLossError: file is too short to be an sstable` when calling `saver.restore()`. It's possible a file is corrupted/truncated.

EDIT: The issue was resolved by copying the .aim file to a temporary directory and unzipping it there.

### DIFF
--- a/aim/artifacts/model.py
+++ b/aim/artifacts/model.py
@@ -3,6 +3,7 @@ import os
 import tempfile
 import zipfile
 from typing import Any
+import shutil
 
 from aim.engine.utils import is_keras_model, is_pytorch_module, is_tensorflow_session, get_module
 from aim.artifacts.artifact import Artifact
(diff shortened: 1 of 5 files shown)


examples/tf_model_load.py
```diff
@@ -0,0 +1,40 @@
...
+    hidden_1 = tf.layers.dense(inputs=input_layer, units=n_hidden_1, activation=tf.nn.relu, name='hidden-1')
+    hidden_2 = tf.layers.dense(inputs=hidden_1, units=n_hidden_2, activation=tf.nn.relu, name='hidden-2')
+    hidden_3 = tf.layers.dense(inputs=hidden_2, units = n_hidden_3, activation=tf.nn.relu, name='hidden-3')
+
+    # Output fully connected layer with a neuron for each class
+    out_layer = tf.layers.dense(inputs=hidden_3, units=num_classes, name='output')
+
+    return out_layer
+
+MODEL_PATH = ''
```
gorarakelyan, line 38: Is model loading tested?

aim/artifacts/model.py
```diff
@@ -23,6 +25,11 @@ def load(path: str) -> (bool, Any):
             model_path = os.path.join(working_dir, path)
             if not os.path.isfile(model_path):
                 return False, None
+        
+        # Create temporary directory
+        dir_name = 'tmp_model'
```
gorarakelyan, line 30: It is better to use `NamedTemporaryFile` or `TemporaryDirectory` with tempfile package

examples/tf_model_load.py
gorarakelyan: Does it also properly work in case of NN built with tf ops?

bhaveshk658: Yes, it works for a NN built with tf ops as well as one built with tf.layers. The path is empty following the examples for PyTorch and Keras because the path can vary depending on where the file is being run from.

aim/artifacts/model.py
bhaveshk658: Got it! Changed to use TemporaryDirectory instead.

PR #417, opened 2021-03-30
Fix broken link

gorarakelyan:
- Remove broken link
- Add experiment description in Session section


gorarakelyan:
Resolves #417

PR #458, opened 2021-05-05
Update keras implementation

### LINKED ISSUE #461: Keras Integration implementation update
Currently, our Keras's Callback implementation only allows passing already made sessions. Would be great to have a choice to only give the Callback `repo` and `name` arguments and let it make the session.

Khazhak:
Allowed Keras users to use repo/experiment arguments instead of creating a session and passing it to the callback.
Closes #461.

### DIFF
--- a/aim/sdk/adapters/keras.py
+++ b/aim/sdk/adapters/keras.py
@@ -27,10 +27,11 @@ class AimCallback(object):
         return cls.__keras_tracker_callback_cls
 
     @classmethod
-    def metrics(cls, session: Optional[Session] = None,
-                repo: Optional[str] = None, experiment: Optional[str] = None):
+    def metrics(cls, repo: Optional[str] = None,
+                experiment: Optional[str] = None,
+                session: Optional[Session] = None):
         # Keep `metrics` method for backward compatibility
-        return cls(session, repo, experiment)
+        return cls(repo, experiment, session)
 
     def __init__(self, repo: Optional[str] = None,
                  experiment: Optional[str] = None,
--- a/aim/sdk/adapters/keras.py
+++ b/aim/sdk/adapters/keras.py
@@ -27,11 +27,13 @@ class AimCallback(object):
         return cls.__keras_tracker_callback_cls
 
     @classmethod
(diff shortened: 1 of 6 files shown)


aim/sdk/adapters/keras.py
```diff
@@ -27,11 +27,13 @@ def __get_callback_cls(cls):
...
     @classmethod
-    def metrics(cls, session: Optional[Session] = None):
+    def metrics(cls, session: Optional[Session] = None,
+                repo: Optional[str] = None, experiment: Optional[str] = None):
         # Keep `metrics` method for backward compatibility
-        return cls(session)
+        return cls(session, repo, experiment)
 
-    def __init__(self, session: Optional[Session] = None):
+    def __init__(self, session: Optional[Session] = None,
```
gorarakelyan, line 35: Let's move session argument to the end of the list, because it will be depreciated soon.

aim/sdk/adapters/keras_mixins.py
```diff
@@ -47,9 +47,23 @@ def _log_epoch_metrics(self, epoch, logs):
...
-        def __init__(self, session: Optional[Session] = None):
+        def __init__(self, session: Optional[Session] = None,
+                     repo: Optional[Session] = None,
+                     experiment: Optional[Session] = None):
             super(KerasTrackerCallback, self).__init__()
-            self._session = session
+
+            if session is None:
+                if repo is None and experiment is None:
+                    print('Neither session nor repo and experiment were passed' +
```
gorarakelyan, line 57: I guess this alert message is redundant here.

aim/sdk/adapters/keras_mixins.py
```diff
@@ -47,9 +47,23 @@ def _log_epoch_metrics(self, epoch, logs):
...
+                          ' to the AimCallback. Will use the default session.')
+                    self._session = Session()
+                else:
+                    self._session = Session(
```
gorarakelyan, line 61: Let's print a warning message here stating for example `passing session object will be deprecated soon, pass repo and experiment arguments explicitly instead`

gorarakelyan (commented):
Left some comments

Issue #479, opened 2021-06-01
Failed tracking due to early session closing 


FrancescoBozzo:
## Reported error
```
Something went wrong in `track`. The process will continue to execute.
`track`: session is closed
`track`: session is closed
`track`: session is closed
`track`: session is closed
```

## System
- OS: Pop!_OS 20.10
- Kernel: 5.11.0-051100-generic
- Docker: version 20.10.6, build 370c289

## How to reproduce
1. Create a `Dockerfile` in the project parent directory:
```Dockerfile
FROM python:3.8
WORKDIR /usr/src/app
COPY requirements.txt ./
RUN pip install --no-cache-dir -r requirements.txt
COPY . .
CMD [ "python", "./test.py" ]
```
2. Create a `test.py` script in the project parent directory:
```python
from aim.sdk.session import Session

sess = Session()
sess.set_params({"foo": "bar"})

for i in range(10):
    sess.track(i, name="val")

sess.close()
```
3. Build and run the container
```bash
docker build -t aim-dev .
docker run -it --rm aim-dev
```

## Hypothesis
I guess the problem is caused by the fact that `self.active` is set to `False` abruptly when the session is closed, even if there are more thread to execute on the `self._queue`.
https://github.com/aimhubio/aim/blob/822c41aaac3d107c9f7a1567717487a25a9f3e10/aim/sdk/session/session.py#L196-L203

https://github.com/aimhubio/aim/blob/822c41aaac3d107c9f7a1567717487a25a9f3e10/aim/sdk/session/session.py#L222-L232

gorarakelyan:
@mihran113 can you please take a look into this? The non-daemon threads(queue jobs) should block the process termination, but somehow the process exits sooner. 
Also, as the queue already runs in a parallel thread and doesn't block ML training, it is worth removing threads and executing queue jobs as normal python functions.

PR #480, opened 2021-06-02
Early session closing

mihran113:
Move the session status setting to non-active in close method to a later stage and change the session's queue jobs from threads to python functions


mihran113:
resolves #479

gorarakelyan (approved):
lgtm

Issue #352, opened 2021-03-13
Ability to save explore views


SGevorg:
By larsr, jaekyeom, mahnerak @ Slack:
Aim explore views are very effective for monitoring and comparing 1000s of experiments at once.
Add an ability to save the explore views so it's easy to navigate back to useful explore states.

PR #481, opened 2021-06-04
Add ability to save explore state/view as bookmark

mihran113:


roubkar:
Resolves https://github.com/aimhubio/aim/issues/352

gorarakelyan (approved):
Great job!

Issue #484, opened 2021-06-15
Hide chart popup tooltip during zooming


roubkar:
### Problem

Chart popup tooltip makes it hard to work with the zoom-in brush.

### Solution

Hide chart popup tooltip when zoom mode is on.

PR #485, opened 2021-06-15
Hide chart popup tooltip during zooming

VkoHov:


roubkar:
Resolves https://github.com/aimhubio/aim/issues/484

roubkar (approved):
lgtm

Issue #502, opened 2021-06-30
The bookmark navigation freezes 


SGevorg:
The new bookmarks feature sometimes doesn't work:
Steps: 
- Go to bookmarks from explore.
- Navigate to a saved bookmark.

Please see the video reproduction of the defect:

https://user-images.githubusercontent.com/3179216/123907582-3b155c00-d987-11eb-8013-2ed0af128ec7.mov

PR #503, opened 2021-06-30
Fix bookmark navigation issue

roubkar:


roubkar:
Resolves https://github.com/aimhubio/aim/issues/502

Issue #504, opened 2021-06-30
Empty metric select on X-axis alignment property change


roubkar:
When the user changes the X-axis alignment option from metric to another one, the metric select field value remains.
It would be great to empty the metric select field after the alignment change.

PR #505, opened 2021-06-30
Empty metric select on X-axis alignment property change

roubkar:


roubkar:
Resolves https://github.com/aimhubio/aim/issues/504

PR #512, opened 2021-07-05
Issues fix/collections abc import 367

fetz236:
Fixed the error regarding potential collections.abc depreciation warnings using a try catch statement. Multiple files altered to protect against Depreciation.


gorarakelyan (approved):
lgtm

PR #514, opened 2021-07-06
Add outliers ignore functionality

arsengit:
- Calculating Y-axis min max values during toggling outliers
- Toggling outliers in LineChart component from Controls


gorarakelyan (approved):
lgtm

roubkar (approved):

PR #515, opened 2021-07-08
Draw LineChart hover attributes

KaroMourad:
Add base mouseover functionality:
  - draw nearest points on hover
  - set nearest points style
  - find closest point on hover
  - set closest point style
  
 Add packages: 
  - bs58check (used by "JS standard code style", popular)
 
 Add hashing functions:
  - contextToHash
  - traceToHash


aim/web/ui_v2/src/components/LineChart/styles.module.css
```diff
@@ -0,0 +1,54 @@
+.HoverLine {
```
roubkar, line 1: @KaroMourad can you please change this file name.
I think `LineChart.module.css` would be more helpful

roubkar (changes requested):
Please change the styles module file name, and resolve conflicts with `main` branch

roubkar (approved):
lgtm

aim/web/ui_v2/src/utils/toHash.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function bufferToString(buffer: Buffer): string {
+  return buffer.toString('utf-8');
+}
+
+export function contextToHash(context: { [key: string]: unknown }): string {
+  return encode(Buffer.from(JSON.stringify(context)));
+}
+
+export function traceToHash({
```
gorarakelyan, line 11: Probably it is better to have one encoder which receives any object(s) and returns encoded string.

gorarakelyan (approved):

aim/web/ui_v2/src/utils/toHash.ts
roubkar: yeah, definitely! @KaroMourad can you please add a separate utility function for encoding and use it in hashing functions?

KaroMourad: Fine , i will fix this

PR #528, opened 2021-07-22
Enhance Metric App Model

arsengit:
- Moved all methods and states from MetricsContainer to metricAppModel
- Changed drawing method of lines in drawLines function
- Fixed bug in updateHoverAttributes function (incorrect attributes while smoothing data)


VkoHov (approved):

aim/web/ui_v2/src/types/utils/d3/drawHoverAttributes.d.ts
```diff
@@ -6,12 +6,12 @@ import HighlightEnum from 'components/HighlightModesPopover/HighlightEnum';
 
 export interface IDrawHoverAttributesProps {
   index: number;
-  data: IProcessedData[];
   visAreaRef: React.MutableRefObject<>;
   attributesNodeRef: React.MutableRefObject<>;
   attributesRef: React.MutableRefObject<{
     xScale?: IGetAxesScale['xScale'];
     yScale?: IGetAxesScale['yScale'];
+    attributesData: IProcessedData[];
```
KaroMourad, line 15: if we don't need this i think we can remove from here `attributesData: IProcessedData[];`

aim/web/ui_v2/src/types/components/LineChart/LineChart.d.ts
```diff
@@ -29,4 +29,11 @@ export interface ILineChartProps {
   axesScaleType: IAxesScaleState;
   highlightMode: HighlightEnum;
   curveInterpolation: CurveEnum;
+  focusedState: {
+    key: string | null;
+    xValue: number | null;
+    yValue: numbae | null;
```
KaroMourad, line 35: Fix `yValue: numbae | null;`

KaroMourad (approved):

aim/web/ui_v2/src/components/HighlightModesPopover/HighlightModesPopover.tsx
```diff
@@ -13,19 +13,19 @@ function HighlightModesPopover(
       <MenuList>
         <MenuItem
           selected={props.mode === HighlightEnum.Off}
-          onClick={props.onChange(HighlightEnum.Off)}
+          onClick={() => props.onChange(HighlightEnum.Off)}
```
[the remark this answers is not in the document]
arsengit: it will cause call stack error

aim/web/ui_v2/src/types/components/LineChart/LineChart.d.ts
arsengit: fixed

roubkar:
@arsengit can you please resolve the conflict with `main` branch

aim/web/ui_v2/src/components/HighlightModesPopover/HighlightModesPopover.tsx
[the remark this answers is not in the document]
KaroMourad: Why?  If the function like this type `onChange: (mode: number) => () => void;` it's mean that the function creates closure and return new function , there is no error if u use it right way

gorarakelyan (approved):

PR #529, opened 2021-07-22
Run metadata API & db schema (draft version)

alberttorosyan:
Draft version of Run metadata storage & API. Covers the following parts:
- sqllite db integration to Repo
- db schema and ORM models
- db migrations infrastructure
- run API (server part only, no SDK support)
- experiments/tags API (no underlying implementation yet)


aim/storage/run_metadata/models.py
```diff
@@ -0,0 +1,82 @@
...
+    Column('run_id', Integer, ForeignKey('run.id'), primary_key=True, nullable=False),
+    Column('tag_id', Integer, ForeignKey('tag.id'), primary_key=True, nullable=False)
+)
+
+
+class Run(Base):
+    __tablename__ = 'run'
+
+    id = Column(Integer, autoincrement=True, primary_key=True)
+    uuid = Column(Text, index=True, unique=True, default=get_uuid)
```
gorarakelyan, line 31: I think we should keep either `uuid` or `hash`. Thoughts?

aim/web/api/__init__.py
```diff
@@ -19,6 +19,15 @@ def create_app():
     from aim.web.api.projects.views import projects_router
     from aim.web.api.tags.views import tags_router
     from aim.web.api.views import general_router
+
+    # v2 API routes
+    from aim.web.api.v2.views.runs import run_router
+    from aim.web.api.v2.views.tags import tag_router
+    from aim.web.api.v2.views.experiments import experiments_router
+
+    app.include_router(run_router, prefix='/api/v2/runs')
```
gorarakelyan, line 28: @mihran113 @alberttorosyan runs used to be named as `commits` in previous versions. Therefore, probably, these blueprints should be merged into one.

aim/web/api/__init__.py
```diff
@@ -19,6 +19,15 @@ def create_app():
...
+    app.include_router(tag_router, prefix='/api/v2/tags')
```
gorarakelyan, line 29: Question: Do we need API versions in URL? 🤔

mihran113: I don't think so, as previous versions won't be supported, no need for versioning

gorarakelyan: Agree

alberttorosyan: @gorarakelyan @mihran113 for the moment we might have both old and new APIs in the same code-base. The existing one has for example endpoint prefix `/api/v1/tags/`, so the new one should be either `/api/v2/tags/` or `/api/tags/` (and all other entities as well). I'm OK with both

aim/storage/run_metadata/models.py
alberttorosyan: the `hash` is effectively used for mapping the sqlite Run entity to the rocksdb data. the `uuid` is somewhat intended for the User (hence URLs constructed based on uuid). It could also be the `run.id`, but in general it's better not to expose  database's internal IDs.  So in this context, you can think of `uuid` as database id. I think we have to keep both.

aim/web/api/__init__.py
gorarakelyan: Actually the new API covers nearly all the functionality of the old one, hence there is no point in keeping v1.

aim/web/api/__init__.py
alberttorosyan: just didn't want to mess with the existing API so the UI team will have time to and keep frontend functional in all common branches. Indeed we should review/drop old APIs once everything is integrated.

aim/web/api/__init__.py
alberttorosyan: @gorarakelyan, then I'll stick to the following path structure: `/api/tags/`, `/api/runs`, etc.

aim/web/api/__init__.py
gorarakelyan: @alberttorosyan Fair enough, but currently UI team is not using any API endpoint. 
All endpoint are going to be refactored, based on the new storage.

aim/web/api/__init__.py
gorarakelyan: @alberttorosyan awesome

aim/storage/run_metadata/models.py
gorarakelyan: Got it. Taking into account that not all the runs would be stored in sqlite, but the ones which metadata was updated(name, tag, experiment, etc), it probably makes more sense to expose `run.hash` to users. The one that always exists and is stored in rocksdb.

PR #539, opened 2021-07-29
Run metadata QL integration

mihran113:


aim/storage/query.py
```diff
@@ -91,6 +96,29 @@ def __call__(
...
+        if self.cache is not None:
+            return self.cache
+
+        runs_factory = self.repo.run_metadata_db.runs()
+        query_results = defaultdict(SafeNone)
+        for run in runs_factory:
+            query_results[run.hash] = {
+                'experiment': run.experiment,
+                'tags': run.tags
+            }
```
alberttorosyan, line 117: @mihran113 I think we have to cache an entire run object, since we might have checks like `run.name == '...'` or `run.description == '...'`

aim/storage/run_metadata/entities.py
```diff
@@ -2,7 +2,14 @@
 from typing import Generic, TypeVar, Iterable, Optional
 
 
+# TODO: [MV] consider making this Singleton
```
alberttorosyan, line 5: 👍

alberttorosyan (commented):
@mihran113 looks good to me. Just need to fix one small thing and can merge PR.

gorarakelyan (approved):
lgtm

PR #544, opened 2021-07-31
Add HighPlot axes brush functionality

VkoHov:
- Add drawParallelAxesBrush method for HiPlot  component
- Filter HiPlot data based on brush selections
- Improve drawParallelHoverAttributes method to work with filtered data
- Improve drawParallelLines  method to work with filtered data


roubkar:
@VkoHov there is an issue with resetting brush selection when there are brushes on multiple axes.

Steps to reproduce:

1. Apply brush selection on multiple axes.
2. Click on one of the axes with a brush selection outside of the selection

Expected output:

The brush is removed from that axis, lines are filtered regarding brush selection on the rest of the axes.

Actual result:

The brush is removed from that axis, lines are not filtered (getting back to initial state/no brush selection applied).

roubkar (approved):

aim/web/ui_v2/src/utils/d3/drawParallelAxesBrush.ts
```diff
@@ -0,0 +1,146 @@
...
+  brushRef,
+  plotBoxRef,
+  plotNodeRef,
+  dimensions,
+  data,
+  linesRef,
+  attributesRef,
+}: IDrawParallelAxesBrushBrushProps): void {
+  brushRef.current.xScale = attributesRef.current.xScale;
+  brushRef.current.yScale = { ...attributesRef.current.yScale };
```
KaroMourad, line 22: `brushRef.current.xScale = attributesRef.current.xScale;`

`brushRef.current.yScale = { ...attributesRef.current.yScale };`
  
  why you need to clone yScale function and whats the difference with xScale , can you explain it please ?

VkoHov: There is a difference between yScale of Highplot and yScale of LineChart. In HighPlot yScale is an array with multiple yScales. So if I did it without spreading it would have the same reference and we would have a problem with the initial yScales.

KaroMourad: Thanks

KaroMourad (approved):

gorarakelyan (approved):

PR #547, opened 2021-08-02
Fix broken functionalities for 'pre_release_v3'

mihran113:
Fix broken functionalities after recent changes on 'pre_release_v3' branch and cast epochs and iters to 'float64' before encoding


gorarakelyan (approved):
lgtm

PR #548, opened 2021-08-02
Add ability switch on/off grouping persistence state

arsengit:
- Added ability to switch on/off grouping persistence state.
- Fixed className naming (BEM) in grouping popovers
- Added visibility state icons in grouping items.


VkoHov (approved):

aim/web/ui_v2/src/services/models/metrics/metricsAppModel.ts
```diff
@@ -189,8 +259,8 @@ function groupData(data: IMetric[]): IMetricsCollection[] {
     }
   }
 
-  let colorIndex = 0;
-  let dasharrayIndex = 0;
+  let colorIndex: any = 0;
```
roubkar, line 262: Why should we use `any` here?

roubkar (changes requested):

roubkar (approved):
Nice job!

PR #550, opened 2021-08-03
Implement set active line method

KaroMourad:
Add: 
 - Implement setActiveLine
 
 Fix:
 - bug in updateHoverAttributes (clear other charts attributes)


VkoHov (approved):

aim/web/ui_v2/src/utils/d3/drawHoverAttributes.ts
```diff
@@ -355,117 +377,93 @@ function drawHoverAttributes(props: IDrawHoverAttributesProps): void {
...
+    force: boolean = false,
+  ): IActivePoint {
+    const { mouseX, mouseY } = getCoordinates({
+      mouse: mousePos,
+      margin,
+      xScale: attributesRef.current.xScale,
+      yScale: attributesRef.current.yScale,
+    });
+    const nearestCircles = getNearestCircles(mouseX);
+    const closestC = getClosestCircle(nearestCircles, mouseX, mouseY);
```
roubkar, line 435: @KaroMourad maybe it would be better to give this variable full name (`closestCircle`), what do you think?

roubkar (approved):
Looks great! 👍

aim/web/ui_v2/src/utils/d3/drawHoverAttributes.ts
KaroMourad: i think you right it would be more readable, i will fix it now

PR #551, opened 2021-08-03
storage/SDK separation phase I:

alberttorosyan:
* moved SDK classes to storage/sdk module
* renamed run_metadata to structured
* moved SafeNone to types.py

Need to drop `query.py` dependency from SDK classes (Repo & Run). This can be done only after merging Run and SQLLite Run objects.


alberttorosyan:
@mihran113 @gorarakelyan this PR gets the job done only partially. But this part is required to do Run objects unification, so it would be good to merge as it is and do the rest afterwards.

mihran113 (approved):

gorarakelyan (approved):
lgtm

PR #552, opened 2021-08-03
Add group option additional fields

arsengit:
- Added additional select options for grouping (group by metric name, experiment name, run hash, context?) 
- Changed grouping selectOptions items type

### DIFF
--- a/aim/web/ui_v2/src/components/ChartPanel/ChartPanel.tsx
+++ b/aim/web/ui_v2/src/components/ChartPanel/ChartPanel.tsx
@@ -98,6 +98,8 @@ const ChartPanel = React.forwardRef(function ChartPanel(
             {props.data.map((lineChartData, index) => (
               <Grid
                 key={index}
+                //TODO generate new only when chart changes
+                // key={lineChartData.map((line) => line.key).join('_')}
                 item
                 xs={
                   props.data.length > 9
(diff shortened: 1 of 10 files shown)


aim/web/ui_v2/src/types/services/models/metrics/metricsAppModel.d.ts
```diff
@@ -116,3 +116,8 @@ export interface IGetPersistIndex {
 }
 
 export type groupNames = 'color' | 'style' | 'chart';
```
roubkar, line 118: @aimhubio/uidevs Let's start types names with uppercase and us type keyword at the end.
e.g.
```js
export type GroupNameType = 'color' | 'style' | 'chart';

export type GroupingSelectOptionType = { ... }; 
```

aim/web/ui_v2/src/components/ChartPanel/ChartPanel.tsx
```diff
@@ -98,6 +98,8 @@ const ChartPanel = React.forwardRef(function ChartPanel(
             {props.data.map((lineChartData, index) => (
               <Grid
                 key={index}
```
roubkar, line 100: shouldn't we change the key logic to avoid drawing an additional first chart?

roubkar (changes requested):

aim/web/ui_v2/src/components/ChartPanel/ChartPanel.tsx
arsengit: Yes We Are, but we need to rerender LineChart when chart grouping state changes. So with the logic we wrote LineChart also rerenders when we changing style and color grouping state. I wrote TODO for this issue.

roubkar (approved):

KaroMourad (approved):
lgtm

PR #560, opened 2021-08-05
Fix linechart active point on zoom mode

KaroMourad:
Fix: 
 - Brush `active` or `focused` point location on zoom
 
Add: 
 - `updateFocusedChart` function can call with an existing `active` or `focused` point(can call without receiving mouse position)

### DIFF
--- a/aim/web/ui_v2/src/types/components/LineChart/LineChart.d.ts
+++ b/aim/web/ui_v2/src/types/components/LineChart/LineChart.d.ts
@@ -47,6 +47,6 @@ export interface IAttributesRef {
   updateScales?: (xScale: IGetAxisScale, yScale: IGetAxisScale) => void;
   setActiveLine?: (lineKey: string) => void;
   updateHoverAttributes?: (xValue: number) => void;
-  updateFocusedChart?: (props?: IUpdateFocusedChartProps) => void;
+  updateFocusedChart?: (params?: IUpdateFocusedChartProps) => void;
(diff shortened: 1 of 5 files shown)


roubkar (approved):

aim/web/ui_v2/src/types/components/LineChart/LineChart.d.ts
```diff
@@ -41,6 +47,6 @@ export interface IAttributesRef {
   updateScales?: (xScale: IGetAxisScale, yScale: IGetAxisScale) => void;
   setActiveLine?: (lineKey: string) => void;
   updateHoverAttributes?: (xValue: number) => void;
-  updateFocusedChart?: (mousePos: [number, number]) => IActivePoint;
+  updateFocusedChart?: (props?: IUpdateFocusedChartProps) => void;
```
arsengit, line 50: lets write function arguments as params instead of props

VkoHov (approved):

aim/web/ui_v2/src/types/components/LineChart/LineChart.d.ts
KaroMourad: fine i will fix this , thanks

KaroMourad: for only simple functions we decide to do yes ? for react components we can use props as well

arsengit (approved):
great

PR #563, opened 2021-08-05
Add bookmarks page model

arsengit:
- Added bookmarks page model
- Deleted bookmarks mock data and mapping bookmarks with fetched data from api


VkoHov (approved):

roubkar:
@arsengit let's keep `IAppData` (app state) structure the same as our current `config` structure

roubkar (approved):

PR #565, opened 2021-08-06
tags/experiments API improvements

alberttorosyan:
* Added run count per tag/experiment to APIs
* Added tag color set/get
* Optimized `ModelMappedColelction len()` method to return query.count() instead of caching all objects


mihran113 (approved):
lgtm

PR #570, opened 2021-08-08
Implement grouped data aggregation logic

roubkar:
- Add aggregation methods constants
- Add a helper for grouped data aggregation calculation
- Apply aggregation calculation on data grouping


KaroMourad (approved):
lgtm

PR #574, opened 2021-08-10
Minimal set on unit-tests for aim storage/SDK/server API

alberttorosyan:


aim/storage/query.py
```diff
@@ -151,7 +151,7 @@ def match(
         metric_name: str = None
     ) -> bool:
 
-        context_proxy = AimObjectProxy(lambda: context.to_dict())
+        context_proxy = AimObjectProxy(lambda: context.to_dict() if context else {})
```
mihran113, line 154: do we need this though?  this won't throw an exception here, and in the `_check` method the exception will be caught and logged

mihran113, line 154: Do we need this though?  this won't throw an exception here, and in the `_check` method the exception will be caught and logged

alberttorosyan: Need to double-check; had a test failing because of this.

gorarakelyan (approved):
lgtm

aim/storage/query.py
alberttorosyan: @mihran113 checked and fixed. The issue was with the logging in `except` block.

mihran113 (approved):
nothing to add, everything's great

PR #584, opened 2021-08-13
Add ability to display notifications on actions

arsengit:
- Added NotificationContainer component
- Added notification add and delete logic
- Added BookmarksNotificationEnum
- Added logic of fetching app config data from back and setting it to metricAppModel


KaroMourad (approved):
lgtm

PR #595, opened 2021-08-16
Faster performance with Auto-Indexing and other tricks

mahnerak:
This PR introduces:
* indexing, union db, index refreshing, shadowing,
* cython implementations for encoders / decoders,
* tuned options for aimrocks,
* generalized PrefixView,
* lazier query variables.


aim/storage/hashing/c_hash.pyxbld
```diff
@@ -0,0 +1,5 @@
+from Cython.Build import cythonize
```
mihran113, line 1: @mahnerak  do we need this file to stay?

aim/storage/query.py
```diff
@@ -50,41 +50,31 @@ def safer_getattr(object, name, default=None, getattr=getattr):
...
 
 
 restricted_globals = {
-    '__builtins__': builtins,
-    '_write_': full_write_guard,
-    '_getiter_': iter,
-    '_getitem_': default_guarded_getitem,
-    '_iter_unpack_sequence_': guarded_iter_unpack_sequence,
-    '_unpack_sequence_': guarded_unpack_sequence
+    "__builtins__": builtins,
```
mihran113, line 65: single quotes, please? 🙏

aim/storage/sdk/repo.py
```diff
@@ -112,17 +113,14 @@ def request(
...
-            prefix = b''
-
-            container_view = SingleContainerView(container=container, read_only=read_only, prefix=prefix)
+            container_view = container
             self.container_view_pool[container_config] = container_view
 
         return container_view
 
-    def iter_runs(self) -> Iterator['Run']:
+    def iter_runs(self) -> Iterator["Run"]:
```
mihran113, line 121: ' 🙏

aim/storage/treeutils_.pyx
```diff
@@ -0,0 +1,263 @@
...
+                yield from unfold_tree(val, path=path + (idx,), unfold_array=unfold_array, depth=depth)
+    elif isinstance(obj, dict):
+        # yield path, ObjectFlag
+        for key, val in obj.items():
+            yield from unfold_tree(val, path=path + (key,), unfold_array=unfold_array, depth=depth)
+    else:
+        raise NotImplementedError
+
+
+cpdef val_to_node(
```
mihran113, line 45: maybe def? as no compile time types are specified

requirements.txt
```diff
@@ -13,6 +13,8 @@ numpy
 wrapt
 cython
 tqdm
+filelock
```
mihran113, line 16: can't see any usage of `filelock` package, is it needed?

mahnerak: Here: https://github.com/aimhubio/aim/blob/6fb8e1d1f31d233cf6541e8233ce847a8e7b010f/aim/storage/container.py#L1-L6

aim/storage/hashing/c_hash.pyxbld
mahnerak: Yes, this is needed for quick development. Makes it easier to use with pyximport.

requirements.txt
mihran113: my bad, have already marked the file as reviewed, that's why couldn't find the usage

mihran113: then we need to add those dependencies also in `setupy.py` as well, forgot to mention that

aim/storage/hashing/c_hash.pyxbld
mihran113: I mean the `c_hash` won't be changed a lot, could be understandable for `treeutils_`

aim/storage/container.py
```diff
@@ -27,13 +29,28 @@ class Container(ContainerView):
...
         self.read_only = read_only
         self._db_opts = dict(
             create_if_missing=True,
-            paranoid_checks=False
+            paranoid_checks=False,
+            keep_log_file_num=10,
+            skip_stats_update_on_db_open=True,
+            skip_checking_sst_file_sizes_on_db_open=True,
+            max_open_files=-1,
+            write_buffer_size = 67108864, # 64MB
```
alberttorosyan, line 44: `64*1024*1024` would be self-explanatory

aim/storage/query.py
```diff
@@ -116,13 +106,18 @@ def __call__(
...
-    return namespace['check']
+                                   mode='eval')
+    return byte_code
+
+class LazyLocals(dict):
+    def __getitem__(self, key: str):
+        try:
+            return super().__getitem__(key)
+        except KeyError:
+            return self['fallback'][key]
```
alberttorosyan, line 120: should we handle the case when there's no 'fallback' key? I do understand that this is used in the scope of this file, so feel free to skip.

aim/storage/sdk/repo.py
```diff
@@ -35,14 +35,15 @@ def __init__(self, path: str, *, read_only: bool = None):
         self.container_view_pool: Dict[ContainerConfig, ContainerView] = WeakValueDictionary()
 
         os.makedirs(self.path, exist_ok=True)
-        # os.makedirs(os.path.join(self.path, 'chunks'), exist_ok=True)
-        # os.makedirs(os.path.join(self.path, 'locks'), exist_ok=True)
-        # os.makedirs(os.path.join(self.path, 'progress'), exist_ok=True)
 
-        self.meta_tree = self.request('meta', read_only=True, from_union=True).tree().view('meta')
+        self._meta_tree = None
```
alberttorosyan, line 39: is `_meta_tree` used?

alberttorosyan (commented):
@mahnerak jan, a couple of small comments here.

aim/storage/sdk/repo.py
mahnerak: `meta_tree` is property right now, to make it lazier.

aim/storage/query.py
mahnerak: If there is no fallback object to look for the key, we might want to raise KeyError (as our locals do not possess the key).
When we return self['fallback'][key] it will automatically raises KeyError as there is no 'fallback' in LazyLocals.

aim/storage/sdk/repo.py
alberttorosyan: I mean _meta_tree. the one with _. I cannot find any usage for it.

aim/storage/query.py
alberttorosyan: got it, thanks!

aim/storage/container.py
alberttorosyan: same for `target_file_size_base` and `max_bytes_for_level_base` (512 MB)

mihran113 (approved):
👏

PR #601, opened 2021-08-18
small fix for tree encoding

alberttorosyan:
fixed case when empty dict is set as value (f.e. trace context, run param)


mahnerak (approved):
We need to inspect this case for Union DB though.

PR #602, opened 2021-08-18
Streaming APIs optimizations

mihran113:
Optimize streaming apis( /runs/search/run/, /runs/search/metric, /runs/search/metric/align/) to start streaming at the beginning of a request instead of processing the whole data and then starting the streaming


alberttorosyan (approved):
looks good

gorarakelyan (approved):

PR #603, opened 2021-08-18
Feature/legacy sdk cleanup

alberttorosyan:
* Remove legacy AimRepo (keep only stuff required for repo convert CLI) 
* Remove legacy CLI commands (aim expriment, etc.) update remaining ones to use new Repo 
* Update aim.sdk.Session class to use new Repo 
* aim.storage enhancements (run step auto-increment, run hash generation, repo path handling, etc. )


aim/sdk/__init__.py
```diff
@@ -1,18 +1,4 @@
 # Main functions
 from aim.sdk.track import track, set_params
 from aim.sdk.flush import flush
-from aim.sdk.init import init
```
gorarakelyan, line 4: `init` method should be kept, it used to create a default session for forwarding `aim.track` to `default_session.track`.

gorarakelyan: It was also used for specifying experiment name and repo path of default session.

aim/sdk/session/session.py
```diff
@@ -2,79 +2,42 @@
...
-    Job = namedtuple('Job', 'target, args, kwargs')
+class Session:
+    sessions = {}
 
     _are_exit_listeners_set = False
     _original_sigint_handler = None
     _original_sigterm_handler = None
 
     def __init__(self, repo: Optional[str] = None,
                  experiment: Optional[str] = None,
```
gorarakelyan, line 24: Add deprecation warning.

aim/engine/utils.py
```diff
@@ -1,311 +0,0 @@
...
-    if is_numpy_number(value):
-        return True
-
-    if is_pytorch_tensor(value):
-        return True
-
-    return False
-
-
-def convert_to_py_number(value):
```
gorarakelyan, line 247: if I am not wrong, this function is used in `sdk/adapters` package.

gorarakelyan (commented):
what do you think about moving `aim/storage/sdk` to `aim/sdk`?

gorarakelyan:
The rest seems good! 👍

gorarakelyan:
+ resource tracker should be initiated in `Run` as well, it tracks system resource usage

aim/sdk/__init__.py
```diff
@@ -1,18 +1,4 @@
...
-from aim.sdk.select import select_metrics, select_runs
```
gorarakelyan, line 5: forward to runs and metrics new search interfaces? +add deprecation warnings

aim/sdk/session/session.py
alberttorosyan: sure

aim/engine/utils.py
alberttorosyan: will double-check this one

aim/sdk/__init__.py
alberttorosyan: should we stub `init` method or should we keep this behavior? as of now, `Repo` (and legacy `Session`) do not require special initialization before usage.  the only thing required is `aim init` being ran for that repo.

aim/sdk/__init__.py
alberttorosyan: > what do you think about moving `aim/storage/sdk` to `aim/sdk`?

totally makes sense. but only after current `aim/sdk` contents are finally dropped.

examples/keras_track.py
```diff
@@ -1,6 +1,6 @@
 from aim.keras import AimCallback
 
-import keras
+from tensorflow import keras
```
gorarakelyan, line 3: This line should remain as `import keras`, because before tf2 keras was a separate high-level ML framework.

gorarakelyan (approved):
looks great

mihran113 (approved):
fantastic job!
https://github.com/aimhubio/aim/pull/603#discussion_r692867563
just noticed you didn't address this comment by Gor, everything else is good.

examples/keras_track.py
alberttorosyan: 👌

PR #614, opened 2021-08-21
Avoid null-terminating C-style strings

mahnerak:
This PR fixes issues with encoded values being truncated with null-byte because of c-style strings are used by cython.


alberttorosyan (approved):
@mahnerak jan, please include `rm aim-playground` in this PR.

alberttorosyan (approved):
lgtm

PR #615, opened 2021-08-21
add sqlite Run every time new Run is started

alberttorosyan:
Removed all the logic associated with the case when sqlite Run is missing. Added `run.experiment = 'default'` for newly created runs

### DIFF
--- a/aim/sdk/errors.py
+++ b/aim/sdk/errors.py
@@ -1,0 +1,2 @@
+class RepoIntegrityError(RuntimeError):
+    pass
--- a/aim/storage/sdk/run.py
+++ b/aim/storage/sdk/run.py
@@ -177,6 +177,7 @@ class Run:
                     raise RepoIntegrityError(f'Missing props for Run {self.hashname}')
                 else:
                     self._props = sdb.create_run(self.hashname)
+                    self._props.experiment = 'default'
 
     def proxy(self):
         run = RunView(self)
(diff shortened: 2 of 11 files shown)


alberttorosyan:
Currently the implementation of `run.props` assumes that sqlite Run model is created first time Run object is initialized (in write mode). Since unit-tests truncate sqlite db every time per test-suite, this breaks all tests run (though 1 test-suite passes). Need to figure-out the way to fix tests without altering this logic.

mihran113:
> Currently the implementation of `run.props` assumes that sqlite Run model is created first time Run object is initialized (in write mode). Since unit-tests truncate sqlite db every time per test-suite, this breaks all tests run (though 1 test-suite passes). Need to figure-out the way to fix tests without altering this logic.

Maybe we need to truncate rocksdb data as well, and create new runs for each test case as well? To provide total independence for each test case?

alberttorosyan:
> > Currently the implementation of `run.props` assumes that sqlite Run model is created first time Run object is initialized (in write mode). Since unit-tests truncate sqlite db every time per test-suite, this breaks all tests run (though 1 test-suite passes). Need to figure-out the way to fix tests without altering this logic.
> 
> Maybe we need to truncate rocksdb data as well, and create new runs for each test case as well? To provide total independence for each test case?

Will definelty get to that at some point.

mihran113 (approved):
looks good, we'll take care of the unit tests after

PR #619, opened 2021-08-22
Update QL aliases and SDK Run interface

mihran113:
-run is available on top level as well as under metric.run
-context is available on metric.context and everything inside context is accessible directly from metric.*
-run params are accessible from top level as a fallback(discuss if need to be removed)


aim/sdk/trace.py
```diff
@@ -240,10 +294,11 @@ def iter(
             if not self.query_traces:
                 statement = True
             else:
+                run_view = run.view()
+                metric_view = MetricView(metric_name, ctx.to_dict(), run_view)
```
alberttorosyan, line 298: maybe it's better to create `ContextView` here and pass to `MetricView`?

aim/storage/query.py
```diff
@@ -133,13 +129,11 @@ def __init__(
...
-        metric_name
+        metric
     ):
+        #TODO: [MV] discuss if run params are needed as a fallback
         namespace = LazyLocals(run=run,
-                               context=context,
-                               ctx=context,
-                               metric_name=metric_name,
+                               metric=metric,
                                fallback=run, **restricted_globals)
```
alberttorosyan, line 137: `fallback` is not needed anymore.

aim/sdk/trace.py
```diff
@@ -163,6 +164,59 @@ def dataframe(
...
+
+
+class MetricView:
+    def __init__(self, name: str, context: dict, run_view: 'RunView'):
+        self.name = name
+        self.run_view = run_view
+        self.context = context
+
+    def __getitem__(self, key):
+        return self.context[key]
```
alberttorosyan, line 198: check for `metric['run']` and `metric['context]`.

alberttorosyan (commented):
A couple fo small comments. The rest looks great.

mahnerak (commented):
`set_params` is kind of confusing. It may mislead users to store trainable parameters instead of hyperparameters or configs.

aim/cli/upgrade/utils.py
```diff
@@ -44,15 +44,15 @@ def convert_run(lrun: LegacyRun, repo: Repo, legacy_run_map, skip_failed):
         lrun.open_storage()
         if lrun.params.get(AIM_MAP_METRICS_KEYWORD):
             del lrun.params[AIM_MAP_METRICS_KEYWORD]  # set internally. no need to copy
-        run[...] = lrun.params
-        run['v2_params'] = {'run_hash': lrun.run_hash}
+        run.set_params(..., lrun.params)
```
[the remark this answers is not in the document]
alberttorosyan: @mahnerak I'm open to suggestions. But since we couldn't find anything better, we decided to leave as `set_params`. At least this is consistent with the `Session` and users are familiar it. In the future we do plan to have dict-like set interface, so maybe no need to introduce new names? @gorarakelyan any thoughts?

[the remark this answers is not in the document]
gorarakelyan: Agree with @alberttorosyan

aim/cli/upgrade/utils.py
```diff
@@ -47,12 +47,12 @@ def convert_run(lrun: LegacyRun, repo: Repo, legacy_run_map, skip_failed):
         run[...] = lrun.params
         run['v2_params'] = {'run_hash': lrun.run_hash}
```
mahnerak, line 48: What is `v2_params` exactly? If you mean some legacy params we do not store anymore - maybe use 'legacy_params'?

mahnerak: v2 seems "newer" at the first glance. Also, these params may have come from v1 too.

mahnerak (approved):

alberttorosyan (approved):
lgtm

aim/cli/upgrade/utils.py
alberttorosyan: > What is `v2_params` exactly? If you mean some legacy params we do not store anymore - maybe use 'legacy_params'?

good point. @gorarakelyan any objections on renaming to `legacy_params`?

PR #644, opened 2021-08-31
Make changes in backend API:

mihran113:
- add tag hard delete
 - add type for dashboard apps and make app state schema-less


alberttorosyan (approved):
looks good

PR #656, opened 2021-09-01
Added run.end_time to web APIs

alberttorosyan:
added `finalized_at` column to Run model and alembic migration.
updated web APIs to return run.end_time


mihran113 (approved):
looks good

PR #683, opened 2021-09-08
Add configs for run_metadata.sqlite

mihran113:
Make run_metadata db configurable depending on which mode "aim" is being used('prod', 'dev' or 'test')


tests/conftest.py
```diff
@@ -29,7 +29,7 @@ def pytest_sessionstart(session):
     Project.set_repo_path(TEST_REPO_PATH)
 
     _init_test_repo()
-    os.environ[AIM_WEB_ENV_KEY] = 'test'
+    os.environ[AIM_ENV_MODE_KEY] = 'test'
```
alberttorosyan, line 32: I think we should set environment before `_init_test_repo()` to make sure run metadata db log is captured.

alberttorosyan (approved):
@mihran113, just one small comment on unit-tests. The rest looks great!

tests/conftest.py
mihran113: done ✅

PR #695, opened 2021-09-09
Add table top bar popover actions

roubkar:
- Implement columns reordering
- Implement rows reordering
- Implement all rows visibility toggling


KaroMourad (approved):
lgtm

PR #714, opened 2021-09-13
Fix run finalization

mihran113:
initialize container's writable db when container object is created, instead of first write access

### DIFF
--- a/aim/storage/container.py
+++ b/aim/storage/container.py
@@ -65,7 +65,7 @@ class Container(ContainerView):
         self._lock_path: Optional[Path] = None
         self._progress_path: Optional[Path] = None
         if not self.read_only:
-            self.preload()
+            self.writable_db
         # TODO check if Containers are reopenable
 
     @property


alberttorosyan (commented):
lgtm

alberttorosyan (approved):

aim/storage/container.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         self._lock_path: Optional[Path] = None
         self._progress_path: Optional[Path] = None
         if not self.read_only:
-            self.preload()
+            self.writable_db
```
mahnerak, line 68: There are some writable-specific parts in `Container.db()`
https://github.com/aimhubio/aim/blob/b7432d995c7c3d8e77f684aa81fc8742fd290189/aim/storage/container.py#L80-L84
Maybe we need to separate writable and read-only Container init in near future?

mahnerak (approved):

aim/storage/container.py
mihran113: Sounds reasonable, we can discuss it later

PR #717, opened 2021-09-13
Implement table resize state change functionality

arsengit:
- Implemented table resize state change functionality(Metric/Params)
- Visualize pending state when calling align by metric endpoint


aim/web/ui_v2/src/pages/Metrics/Metrics.scss
```diff
@@ -31,40 +29,18 @@
...
-    }
-  }
 
   .Metrics__chart__container {
-    -webkit-box-flex: 0.5;
-    -ms-flex: 0.5 1 0;
-    flex: 0.5 1 0;
     min-height: 372px;
+    &__fullHeight {
+      flex: 1 100% !important;
```
KaroMourad, line 36: Is this necessary to write '!important' here ? Lets write without '!important' if we can

arsengit: Yes it's necessary, because it should override inline style.

KaroMourad (approved):

PR #718, opened 2021-09-14
Recover scroll position on new data load

roubkar:
Fixed Runs table unmount-mount issue caused by `key` prop value change.


rubenaprikyan (approved):
Great !

PR #719, opened 2021-09-14
Add ability to toggle column visibility in Runs table

roubkar:
- Add ability to toggle visibility of a single column
- Add ability to hide/show all columns
- Add ability to hide columns with the same value for each row (show table diff)


rubenaprikyan (approved):
Thank's for doing this!

PR #724, opened 2021-09-15
Fix Table export CSV button for all (Metrics, Params, Runs) pages

KaroMourad:
Fix:
 - sorted by order
 - filter hidden columns
 - 'getDataAsTableRows' function
 
 Add: 
 - xStep value for table columns for value, epoch, timestamp
 - tune grouped data export for table


VkoHov (approved):

rubenaprikyan (approved):
Great!

PR #738, opened 2021-09-16
Fix handling bug of api error corner cases

rubenaprikyan:
- Fix handling bug of api error corner cases
- change memo's comparator logic to re-render table after changing rowHeight prop


roubkar (approved):

rubenaprikyan:
- Fix handling bug of api error corner cases
- Change memo's comparator logic to re-render table after changing rowHeight prop

PR #745, opened 2021-09-16
Add group aggregation data to table value column

roubkar:
- Modify value column content for group header rows
- Apply group aggregation data change to table dynamic update function


arsengit (approved):
nice

PR #747, opened 2021-09-16
Update rendering dependencies in Table component

arsengit:
Updated rendering dependencies in Table component
---
- Deleted unnecessary key property from Table component (Metrics/Params)


rubenaprikyan (approved):
This is a better solution than setting a key, thanks!

PR #758, opened 2021-09-20
Display group config properties in table

roubkar:
- Add group config elements to table group first column
- Display row color on rows first cell


roubkar:
Screen Shot 2021-09-20 at 17 26 42

arsengit (approved):

PR #761, opened 2021-09-20
Fix chart/svg drawing issue for Safari

arsengit:
Fixed chart/svg width and height properties
---
- Update min height for table and chart components (Metrics/Params)


KaroMourad (approved):
lgtm

PR #766, opened 2021-09-21
Implement store column resized width value

VkoHov:
- Added updateColumnWidth method to paramsAppModel, runsAppModel and metricsAppModel
- Implement column width value storing in Params, Runs and Metrics pages
- Implement reset width functionality


rubenaprikyan (approved):
Looks good!

PR #768, opened 2021-09-21
Rename, Refactor, Remove, Restyle, Rewrite

mahnerak:
This PR introduces to number of changes, including renaming, refactoring of (mainly) `aim.storage` abstractions, removal of obsolete code, fixing type-compatibilities, removing unnecessary dependencies, moving some implementations to a more reasonable abstraction layer.

Changes include:
* Documentations for `Container`s,
* Renaming Storage base classes (including `Container` -> `RocksContainer`, `ContainerView` -> `Container`, etc.),
* Getting rid of unnecessary `aimrocks` dependencies,
* Separate implementations `delete` and `delete_range`
   (Previously `delete` was actually removing the range and there was no pure `delete` op),
* More generic `set`, `delete`, `delete_range` operations supporting batched writes,
* `delete_range` now supports `[begin, end)`, not just `prefix`-ed ones
   (The previous implementation was featuring some higher-level concepts we wanted to avoid),
* Get value _with defaults_ interface for `Container`s,
* Completely getting rid of `iterlevel`
   (Originally designed to be used by `TreeView.keys()` but `Container.walk()` made it obsolete),
* Fixing type-compatibility of implementations with interfaces,
* Avoiding ugly shortcuts like `E_encode_path`

More modifications are coming in the next PR.


aim/storage/rockscontainer.py
```diff
@@ -0,0 +1,489 @@
...
+    ) -> aimrocks.WriteBatch:
+        """Creates a new batch object to store operations in before executing
+        using :obj:`RocksContainer.commit`.
+
+        The operations :obj:`RocksContainer.set`, :obj:`RocksContainer.delete`,
+        :obj:`RocksContainer.delete_range` are supported.
+
+        See more at :obj:`RocksContainer.commit`
+        """
+        return aimrocks.WriteBatch()
```
alberttorosyan, line 402: here we expose rocksdb-specific object to outer world. maybe it will make sense to return some wrapper instead 🤔 ?

alberttorosyan: I do understand the batch is just passed to set/delete methods and commit, but it's still accessible to "outer world"

mahnerak: Right now we implicitly define the `protocol` in `Container` docs:
https://github.com/aimhubio/aim/blob/cdc0360568af879d52c1e332f1d1ca538ef00f42/aim/storage/container.py#L112-L118
Not sure about wrapping the batch object. Instead, we can define the interface / protocol, and annotate the return type of `Container.batch()` and the argument of `Container.commit()` with that.

aim/sdk/repo.py
```diff
@@ -192,10 +192,10 @@ def _get_container(
         if container is None:
             path = os.path.join(self.path, name)
             if from_union:
-                container = UnionContainer(path, read_only=read_only)
+                container = RocksUnionContainer(path, read_only=read_only)
```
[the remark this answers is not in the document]
mahnerak: @alberttorosyan This is a great idea, (2) especially to prepare the codebase before the stable v3.0.0 possible to (dynamically) use different containers.
For example, to have a more optimized `Container` implementation for `seqs` / `metrics`, and the newer versions will autodetect the container format (instead of using a certain implementation based on the `VERSION`)

As we only touch the SDK as part of our renaming process, let's skip these modifications (2) and refactor them in a different PR.

aim/storage/rockscontainer.py
```diff
@@ -0,0 +1,489 @@
...
+        self._db = None
+        self._lock = None
+        self._wait_if_busy = wait_if_busy  # TODO implement
+        self._lock_path: Optional[Path] = None
+        self._progress_path: Optional[Path] = None
+        if not self.read_only:
+            self.writable_db
+        # TODO check if Containers are reopenable
+
+    @property
```
mihran113, line 63: @mahnerak  I think there's a need to document properties as well, as it would make clear what's the exact difference between  db and writable_db

gorarakelyan (approved):
looks great

aim/sdk/repo.py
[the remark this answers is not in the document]
alberttorosyan: Marking this as resolved. Will have a separate conversation and task for this.

aim/storage/rockscontainer.py
alberttorosyan: same here

aim/storage/rockscontainer.py
mihran113: marking this resolved as well, as @mahnerak assured that this would be done in a separate PR (for rocks specific documentation)

mihran113 (approved):

PR #775, opened 2021-09-22
Sub app division for apis and static files

mihran113:
Use separate sub apps for apis and static-files endpoints to avoid double zipping for static files (only apis sub app will have GZIPMiddleware)


alberttorosyan (approved):
Looks good

PR #777, opened 2021-09-22
Fix chart tooltip positioning

KaroMourad:
Fix: 
 - there are some cases when tooltip hide active point

### DIFF
--- a/aim/web/ui/src/components/ChartPanel/ChartPanel.tsx
+++ b/aim/web/ui/src/components/ChartPanel/ChartPanel.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react';
+import React from 'react';
 import { Grid, PopoverPosition, GridSize, Typography } from '@material-ui/core';
 import _ from 'lodash-es';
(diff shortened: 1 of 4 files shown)


aim/web/ui/src/components/ChartPanel/ChartPanel.tsx
```diff
@@ -17,6 +17,7 @@ import {
 } from 'types/utils/d3/drawHoverAttributes';
 
 import './ChartPanel.scss';
+import getPositionBasedOnOverflow from '../../utils/getPositionBasedOnOverflow';
```
VkoHov: Change it to the absolute path.

VkoHov (approved):

aim/web/ui/src/components/ChartPanel/ChartPanel.tsx
KaroMourad: it unused import i remove it , thanks

VkoHov (approved):

PR #778, opened 2021-09-22
Improve URL sync with the app state

roubkar:
- Add a single `updateURL` function for App `config` state change
- Do not update URL on chart hover
- Add a listener on browser history navigation to re-render the app (back and forward buttons)


VkoHov (approved):

aim/web/ui/src/pages/Metrics/MetricsContainer.tsx
```diff
@@ -90,23 +93,19 @@ function MetricsContainer(): React.FunctionComponentElement<React.ReactNode> {
...
-      metricAppModel.updateChartStateUrl();
-    }
-  }, [metricsData?.config?.chart]);
-
-  React.useEffect(() => {
-    if (metricsData?.config?.select) {
-      metricAppModel.updateSelectStateUrl();
+    if (!!metricsData.config) {
+      if (
+        metricsData.config.grouping !== getStateFromUrl('groupong') ||
```
KaroMourad, line 100: fix typo `getStateFromUrl('groupong')` to `getStateFromUrl('grouping')`

aim/web/ui/src/pages/Params/ParamsContainer.tsx
```diff
@@ -54,23 +56,19 @@ function ParamsContainer(): React.FunctionComponentElement<React.ReactNode> {
...
-      paramsAppModel.updateChartStateUrl();
-    }
-  }, [paramsData?.config?.chart]);
-
-  React.useEffect(() => {
-    if (paramsData?.config?.select) {
-      paramsAppModel.updateSelectStateUrl();
+    if (!!paramsData.config) {
+      if (
+        paramsData.config.grouping !== getStateFromUrl('groupong') ||
```
KaroMourad, line 63: here too , fix typo getStateFromUrl('groupong') to getStateFromUrl('grouping')

aim/web/ui/src/pages/Metrics/MetricsContainer.tsx
```diff
@@ -75,6 +77,7 @@ function MetricsContainer(): React.FunctionComponentElement<React.ReactNode> {
     if (route.params.appId) {
       appRequestRef = metricAppModel.getAppConfigData(route.params.appId);
       appRequestRef.call().then(() => metricAppModel.getMetricsData().call());
+      metricAppModel.setDefaultAppConfigData();
```
KaroMourad, line 80: if we use this method for if/else cases why we need to write it in if/else block, we can take out from this block if i'm not mistaken

roubkar: yeah, you're right, I actually should have put it right after `metricAppModel.getMetricsData().call()` inside promise then

KaroMourad (approved):

PR #780, opened 2021-09-23
Integrate bundle analyzer

rubenaprikyan:
Bundle analyzer tool, to analyze compressed/no-compressed build file sizes, to improve performance.

**Usage**
`$npm run analyze-bundles`
Will open a server on `localhost:8888`

Screen Shot 2021-09-23 at 1 41 33 PM


roubkar (approved):

VkoHov (approved):

arsengit (approved):

KaroMourad (approved):
What do you think do we need to install types for bundle analyzer ? 
@types/webpack-bundle-analyzer

rubenaprikyan:
> What do you think do we need to install types for bundle analyzer ?
> @types/webpack-bundle-analyzer

This file will not compiled as a typescript file
We haven't any usage of this, just a node file running by `node`

To use typescript it need to have  another `tc` building process, and need to create new config for this, I think js is enough at the moment.

KaroMourad (approved):

PR #782, opened 2021-09-23
Remote fs support with sshfs

mihran113:
Add automatically ssh mounting option for repo paths following this format: "ssh://[user:password@[host]:remote_repo_path" with fallbacks to environment variables for user, password, host and identity(ssh_key) file


gorarakelyan:
@mihran113 it would be great to add comment the code

aim/sdk/utils.py
```diff
@@ -1,7 +1,170 @@
...
+import time
 import uuid
-from typing import Union
+from typing import Optional, Tuple, Union
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def get_username():
```
gorarakelyan, line 14: what about moving sshfs utils to `aim/ext/sshfs` directory?

gorarakelyan:
code looks great

aim/sdk/utils.py
mihran113: yup, great advice, done ✅

mihran113:
> @mihran113 it would be great to add comment the code



> @mihran113 it would be great to add comment the code

@gorarakelyan added comments to the code, please have a look

aim/sdk/repo.py
```diff
@@ -54,15 +55,21 @@ def __init__(self, path: str, *, read_only: bool = None, init: bool = False):
...
-        self.path = os.path.join(path, AIM_REPO_NAME)
+        self._mount_root = None
+        if path.startswith('ssh://'):
+            self._mount_root, self.root_path = mount_remote_repo(path)
+        else:
+            self.root_path = path
+        self.path = os.path.join(self.root_path, AIM_REPO_NAME)
 
         if init:
             os.makedirs(self.path, exist_ok=True)
```
alberttorosyan, line 66: does this work for remote fs path?
I think it's better to keep `sefl.path` as the mounted local path, and keep `self._remote_path` when user provides `ssh://` path

alberttorosyan (commented):
@mihran113 jan, looks awsome! just a couple of small comments.

aim/sdk/repo.py
mihran113: we won't need the remote path in any point of the execution, I had `_remote_path` in the initial version, but then decided not to keep it, and just use `_mount_root` for the purpose of checking if we have a remote repo mounted or not, and  `os.makedirs` operation works fine, because `self.root_path` will be already mounted to some local directory

aim/ext/sshfs/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    )
+    child_std_out, _ = child.communicate()
+    sshfs_executable = child_std_out.rstrip()
+    exit_code = child.wait()
+    if exit_code != 0:
+        if platform.system().lower() == 'darwin':
+            raise subprocess.SubprocessError('Could not find sshfs installation. \n'
+                                             'To install sshfs please run the following commands:\n'
+                                             'brew install macfuse \n'
+                                             'brew install brew install gromgit/fuse/sshfs\n'
```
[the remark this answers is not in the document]
mihran113: noted, will remove, thanks for pointing out :)

aim/sdk/repo.py
alberttorosyan: OK, got it. Thanks!

aim/ext/sshfs/utils.py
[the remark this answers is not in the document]
mihran113: done ✅

alberttorosyan (approved):
looks great

PR #786, opened 2021-09-25
Move track into separate thread

alberttorosyan:
Added task queue ext.
Use single thread task queue for executing `Run.track`


aim/ext/task_queue/queue.py
```diff
@@ -0,0 +1,62 @@
...
+            thread = threading.Thread(target=self.worker)
+            thread.daemon = True
+            self._threads.append(thread)
+            thread.start()
+
+    def register_task(self, task_func, *args, **kwargs):
+        if self._stopped:
+            logger.debug('Cannot register task: task queue is stopped.')
+        else:
+            self._queue.put((task_func, args, kwargs))
```
mahnerak, line 31: Let's do two things here:
1. Having a nonzero (though a sufficiently big number) `max_backlog`
2. Warn a user something like:
     > It seems like you are logging too often which may cause performance degradation. Consider logging / tracking every 100 iterations instead.

(1) is not mandatory as we can log such a warning even before `Queue.maxsize` is reached.

mahnerak (changes requested):

aim/sdk/run.py
```diff
@@ -320,7 +321,22 @@ def track(
 
         Appends the tracked value to metric series specified by `name` and `context`.
         """
+
+        # since worker might be lagging behind, we want to log the timestamp of run.track() call,
+        # not the actual implementation execution time.
         track_time = time()
+        self.repo.tracking_queue.register_task(self._track_impl, value, track_time, name, step, epoch, context=context)
```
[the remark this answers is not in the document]
alberttorosyan: This is a great point. Same applies to `context` I guess. I ust concerned about doing this in every case 🤔.

aim/ext/task_queue/queue.py
alberttorosyan: Re: max queue size. I was keeping this in mind; just need to know what we consider as a reasonabe number. Any suggestions?
Re: warning message. How can we know user tracks once every N iterations? Maybe something more generic, like `'Consider tracking at lower pace'`

mahnerak: IDK, maybe 1-10 million? Maybe less, but not more (10 million of plain float writes is already > 100MB, let's not use too much memory, we better block the main thread and log a warning).

Yes, `'Consider tracking at a lower pace'` is much better. Also, what if we have FAQ section in our docs and additionally refer to them (output URLs) for this and other such cases.

aim/sdk/run.py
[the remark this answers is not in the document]
mahnerak: Deep-copying is going to be okay in most of the cases when the data is not too big.
Otherwise, we can think of ways to make such heavy objects to be immutable. The class `Context` was supposed to be one initially, for the same reason.

aim/ext/task_queue/queue.py
alberttorosyan: Yeah, having FAQ section is a great idea. Will add TODO item for that. For now lets use 10mln. as a max size and warn user when queue is 80% full.

mahnerak (approved):

aim/ext/task_queue/queue.py
```diff
@@ -0,0 +1,68 @@
...
+        pending_task_count = self._queue.qsize()
+        if pending_task_count:
+            logger.warning(f'Processing {pending_task_count} pending tasks in queue \'{self.name}\'... '
+                           f'Please do not kill the process.')
+            self._queue.join()
+        self._shutdown = True
+        logger.debug('No pending tasks left.')
+
+    def __del__(self):
+        print('del task queue')
```
mihran113, line 67: forgotten `print`?

mihran113 (commented):
everything else is great, just a minor comment there :)

aim/ext/task_queue/queue.py
alberttorosyan: yep, thanks for pointing

mihran113 (approved):

PR #788, opened 2021-09-25
Move app model methods to seperate files

arsengit:
Separated App model related methods into separated files
---
- Deleted some duplicated functions from metrics, params and runs app models


KaroMourad (approved):
lgtm

PR #799, opened 2021-09-28
change run hash length to 12 symbols (avoid collisions)

alberttorosyan:


mihran113 (approved):

gorarakelyan:
what about increasing the length to 16 chars? or even 24?

gorarakelyan:
@alberttorosyan @mahnerak ^

alberttorosyan:
@gorarakelyan checked with 16 symbols. no difference for up to 10 mln. runs.

gorarakelyan:
btw git hashes are 40 digit strings

>Every time a commit is added to a git repository, a hash string which identifies this commit is generated. This hash is computed with the SHA-1 algorithm and is 160 bits (20 bytes) long. Expressed in hexadecimal notation, such hashes are 40 digit strings.

gorarakelyan:
let's set to 24 if there are no objections @mahnerak @alberttorosyan

mahnerak:
Yes, let's stick with longer hashes, 24 is good.
The only problem is `uuid` returns a digest of 16 chars (the first 2 are fixed).
So maybe need to reimplement the `generate_run_hash`.

alberttorosyan:
`uuid4().hex` is 32 symbols. we can use that in `generate_run_hash`. @mahnerak @gorarakelyan any objections?
and BTW, since this PR is already merged, will open a separate one for that

PR #804, opened 2021-09-28
Fix UI layout shift to top on opening browser devtools from bottom

arsengit:
- Deleted unnecessary `justify-center` style from metrics and params containers
- Added active link detector function in `Sidebar` component
- Overridden `TextField` default styles


roubkar (approved):

rubenaprikyan (approved):
lgtm

PR #807, opened 2021-09-28
Handle displaying of empty context in UI

roubkar:
If the metric has an empty object as a context value we are removing it from Metrics Explorer select form autocomplete.
Also for Runs and Params Explorer we are rendering the `No context` label in the table column header.


KaroMourad (approved):

KaroMourad (approved):
lgtm

PR #809, opened 2021-09-29
Feature/run props caching fixes

alberttorosyan:
cache per repo
fallback to query if run is not found in cache
add missing run to the cache if cache exists (small optimization)
use `structured_db` runs instead of `rocksdb` in `/projects/activity` endpoint


mihran113 (approved):
Looks great, tested myself as well, works as expected :)

mihran113:
@gorarakelyan there's a slight concern here, that when activity api uses only sqlite data, it can show all the runs, but run_meta tree is only eventually consistent, so the runs table sometimes can get behind the numbers of activity api, any comments about this?

PR #813, opened 2021-09-29
Create api caller service using worker

rubenaprikyan:
**Experimental**


roubkar (commented):
@rubenaprikyan the implementation of workerized call looks great! 

Let's make this section a bit compact/smaller like `Bookmark` action in Metrics and Params explorers

Screen Shot 2021-09-30 at 22 18 27


Also, let's store the value of this toggle in the `localStorage` so that it will be persistent.

PR #816, opened 2021-09-30
use current branch for package push workflow

alberttorosyan:
removed hard-coded branch `main` in **python-package** workflow


gorarakelyan (approved):
lgtm

PR #821, opened 2021-10-01
Remote fs and tracking queue fixes

mihran113:
- fix the issue with additional reference to run from tracking thread(this was causing the last run to not finalize) 
- remove `allow_other` option from sshfs options list as on ubuntu it was disabling the auto-mounting and is not really needed by aim


alberttorosyan (approved):
lgtm

PR #823, opened 2021-10-03
Fix the issue with overshadowing

mahnerak:
In Union db we have an issue, let's call it `overshadowing` - when only a part of the given prefix is in the union but it ignores all the other chunks.

This PR fixes the issue when acquiring a view of union db (`PrefixView` in particular) on `chunks`, and it treats index db as a regular chunk db and always redirects the view into the index. The issue lead to undiscoverable `run`s and was able to only detect indexed / finalized ones.

### DIFF
--- a/aim/storage/union.py
+++ b/aim/storage/union.py
@@ -193,7 +193,7 @@ class DB(object):
             self._get_db(prefix, path, self._dbs, new_dbs)
 
         if index_db is not None:
-            new_dbs[index_prefix] = index_db
+            new_dbs[b""] = index_db
         self._dbs = new_dbs
         return new_dbs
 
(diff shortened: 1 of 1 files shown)


aim/storage/union.py
```diff
@@ -205,6 +205,7 @@ def get(self, key: bytes, *args, **kwargs) -> bytes:
             # Shadowing
             if key.startswith(prefix):
                 return db.get(key)
+        return self.dbs[b""].get(key)
         raise ValueError
```
alberttorosyan, line 209: this line should be removed

alberttorosyan (changes requested):

aim/storage/union.py
alberttorosyan, line 209: this line should be removed

alberttorosyan: I don't know why byt Github shows the comment for line 208. What I mean, is that line `209` should be removed.

alberttorosyan (approved):
lgtm

PR #846, opened 2021-10-08
Convert search params into query string in advanced mode

arsengit:
Converting selected search params into ql in advanced mode
---
- Updated Params icon
- Tuned `updateUrl` function use cases in app model
- Added form actions in SelectForm components


arsengit:
> @arsengit, please add a history listener also for the `Runs Explorer` page (see `RunsContainer.tsx`)
resolved

PR #873, opened 2021-10-19
URI changes/service improvements based on discussions with Albert

mihran113:


mihran113:
> lets merge now (to not block UI team) but think about more elegant way of getting Image blob path.

yeah, sure, at least it's working for now 😄

PR #890, opened 2021-10-20
Added sizes dict for Icon buttons

arsengit:
Added `withOnlyIconSize` dict for Icon buttons


roubkar:
@arsengit the icon wrapper element became more narrow, not sure whether it is button padding or width issue

PR #936, opened 2021-10-27
Add performance tests for query and data collection execution times

mihran113:
Query selection is up for a debate, as the owner of the logs @mahnerak can suggest what are the best query options to conduct performance tests on


alberttorosyan (approved):
looks good. just need to add concurrency group to `storage-performance-checks` job to ensure max one job runs at any given time

PR #938, opened 2021-10-28
Implement the lazy loading of images based on their occurrence in viewport

VkoHov:
- Implement the lazy loading of images vertically
- Implement the lazy loading of images horizontally
- Implement getting images blobs functionality only shown in the viewport
Now our Images explore render only images that are shown in the viewport and we wouldn't get images blobs until they wouldn't be in the viewport


KaroMourad (approved):

roubkar (approved):

arsengit (approved):
nice job

Issue #456, opened 2021-05-04
Virtualize table columns


gorarakelyan:
Optimize table mid pane to not render those columns which are out of the user's viewport.

roubkar:
The current implementation of table column virtualization has space for improvements.
Now dummy columns are being rendered without any content (no cells are rendered) outside of the viewport, this improves the performance. However, when there are many dummy columns, the UI still freezes on some interactions.

Reopening this issue to implement a better virtualization mechanism to improve UI rendering performance.

PR #946, opened 2021-11-01
Improve rendering performance by virtualizing table columns

roubkar:
Render column cells only in case if the column is visible within the viewport (+ some offset)


roubkar:
Resolves https://github.com/aimhubio/aim/issues/456

arsengit (approved):

KaroMourad (approved):

PR #947, opened 2021-11-01
Separate `TreeView.view()` and `TreeView.subtree()`

mahnerak:
This PR:
* Introduces `TreeView.subtree()` to obtain a treeview lazily: without requesting the underlying storage,
* Fixes the issue with containers not raising an `KeyError` when the key is absent  and returns `None` instead,
* Defines `TreeView.view()` semi-lazy now (by default, as long as `resolve` is not set). It will request only the type of the object (could be a `TreeView`, a primitive such as `str` and `int`, or a `CustomObject`) and create an object accordingly.


aim/storage/treeview.py
```diff
@@ -42,7 +54,7 @@ def __getitem__(
         self,
         path: Union[AimObjectKey, AimObjectPath]
     ) -> AimObject:
-        ...
+        return self.collect(path)
```
alberttorosyan, line 57: can we also implement `get()` via `__getitem__`?

mahnerak: Thanks, I missed that.

alberttorosyan (approved):
lgtm

Issue #963, opened 2021-11-04
Major performance issue when reading long metrics from large number of experiment


gorarakelyan:
## 🐛 Bug

When metric steps count is greater than 10k it becomes slow to open aimrocks(cython wrapper of rocksdb) storage(~0.5s). 

### To reproduce

Example configuration:
Number of runs: 100
Number of metrics in each run: 10
Number of tracked steps in each metric: 10K

Query execution takes up to 150seconds

### Solution

Speed up aimrocks storage opening.

gorarakelyan:
Resolved in Aim v3.0.5

PR #967, opened 2021-11-05
Flush methods on container close

mihran113:
Added `flush` and `flush_wal` on container close and changed db write options based on previous discussions

### DIFF
--- a/aim/storage/rockscontainer.py
+++ b/aim/storage/rockscontainer.py
@@ -104,21 +104,20 @@ class RocksContainer(Container):
         for k, v in self.items():
             index[k] = v
 
-        self._progress_path.unlink()
-        self._progress_path = None
-
         self._db.flush()
         self._db.flush_wal()
+
+        self._progress_path.unlink()
+        self._progress_path = None
 
     def close(self):
         """Close all the resources."""
         if self._lock is not None:
+            self._db.flush()
+            self._db.flush_wal()
             self._lock.release()
             self._lock = None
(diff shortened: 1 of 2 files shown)


aim/storage/rockscontainer.py
```diff
@@ -106,12 +107,18 @@ def finalize(self, *, index: Container):
...
+        self._db.flush_wal()
+
     def close(self):
         """Close all the resources."""
         if self._lock is not None:
             self._lock.release()
             self._lock = None
         if self._db is not None:
+            if not self.read_only:
+                self._db.flush()
```
alberttorosyan, line 120: I guess we have to keep db locked while flushes are running.

aim/storage/rockscontainer.py
```diff
@@ -106,12 +107,18 @@ def finalize(self, *, index: Container):
         self._progress_path.unlink()
         self._progress_path = None
 
+        self._db.flush()
```
alberttorosyan, line 110: as with the locks, db is considered as 'in progress' unless flushed

alberttorosyan (changes requested):
@mihran113 jan, left comments below

aim/storage/rockscontainer.py
mihran113: yup, legit notices, will change ✅

alberttorosyan (approved):
lgtm

mihran113:
resolves #963

PR #968, opened 2021-11-05
Run rocksdb flush/flush_wal if non-empty wal file was found in readonly mode

alberttorosyan:


alberttorosyan:
Changed the base branch to the `feature/rocksdb-options-tuning`

mahnerak:
Let's not hurry with this. There are plenty of scenarios we should consider before doing things this risky.

For example, what if the fs is provided in read-only mode? What if the user has no permissions of write?

alberttorosyan:
> Let's not hurry with this. There are plenty of scenarios we should consider before doing things this risky.
> 
> For example, what if the fs is provided in read-only mode? What if the user has no permissions of write?

Agree with the part of read-only fs. This should be addressed by just silently skipping runs which cannot be optimized. But we have to do this, since without run proper finalization fix (and graceful shutdown) queries will be too slow. We can revise this part once we're confident that WAL flushes/compactions are done at the end of tracking process.

alberttorosyan:
> Agree with the part of read-only fs. This should be addressed by just silently skipping runs which cannot be optimized. But we have to do this, since without run proper finalization fix (and graceful shutdown) queries will be too slow. We can revise this part once we're confident that WAL flushes/compactions are done at the end of tracking process.

+ we need to fix already generated runs, even if users do not run `aim reindex`

aim/storage/rockscontainer.py
```diff
@@ -487,3 +488,32 @@ def prev_key_value(
...
+        def non_empty_wal():
+            for wal_path in self.path.glob('*.log'):
+                if os.path.getsize(wal_path) > 0:
+                    return True
+            return False
+
+        if non_empty_wal():
+            lock_path = self.prepare_lock_path()
+
+            with FileLock(str(lock_path), timeout=10):
```
mihran113, line 509: @alberttorosyan jan won't this throw an error, if lock couldn't be acquired inside timeout (this would be the case if the run is still actually in progress)? The read operation would fail in this case

mahnerak:
> we need to fix already generated runs, even if users do not run `aim reindex`

We better print a message like _"You have non-clean storage state. Please run `aim reindex` in order to optimize them."_

aim/storage/rockscontainer.py
alberttorosyan: good point. will add exception handling

alberttorosyan:
> > we need to fix already generated runs, even if users do not run `aim reindex`
> 
> We better print a message like _"You have non-clean storage state. Please run `aim reindex` in order to optimize them."_

@mahnerak, it terms of UX it's better to handle runs storage optimization by ourselves (of course without failing the query). This is what we agreed based on discussions with @gorarakelyan.

mihran113 (approved):

alberttorosyan:
resolves #963 : fix sub-optimal storage for queried runs.

PR #985, opened 2021-11-08
Feature/container key value separation

mahnerak:
This PR introduces functionality of storage to store big-value records separately for efficient reads:
- iterating over all the values in a range will not include such big values,
- instead, you find proxy objects capable of loading the contents on-demand.
This is especially useful for such data that is not often loaded from the storage.

For this reasons we implement `BLOB` interface for wrapping primarily `bytes` objects (though it may extend to other primitives as well with a very little effort).
Calling `bytes(blob_instance)` will force load the contents.

The `BLOB` is supported on `Container`-level. In the current implementation of `RocksContainer` we use zero-buffers `b''` as a flag to denote values that have their content separately under `BLOBS\xfe`.

The performance of writes *may be reduced* because of extra puts / deletes.
Consider, first you put a `BLOB` into a key `b'Z'` (which puts a flag on `b'Z'` and contents on `b'BLOBS\xfeZ'`). Then if you want to put even a regular `b'foo'` value on the same `b'Z'` key it has to delete the former `BLOB` contents left on `b'BLOBS\xfeZ'`.


aim/sdk/uri_service.py
```diff
@@ -52,7 +53,10 @@ def request_batch(self, uri_batch: List[str]) -> Iterator[Dict[str, bytes]]:
 
             # TODO: [MV] change to some other implementation of view when available
             #  which won't collect in case of custom objects
-            yield {uri: container.tree().view(resource_path).collect()}
+            data = container.tree().subtree(resource_path).collect()
```
mahnerak, line 56: Why not to use [resource_path] instead?
Also, just by looking at the implementation of `decode_uri()` it looks like `resource_path` is always string :/ Am I correct?

aim/sdk/uri_service.py
```diff
@@ -52,7 +53,10 @@ def request_batch(self, uri_batch: List[str]) -> Iterator[Dict[str, bytes]]:
...
+            if isinstance(data, BLOB):
+                data = data.load()
+            yield {uri: data}
 
         # clear container pool
         self.container_persistent_pool.clear()
```
mahnerak, line 62: If there are two parallel `request_batch()` calls, the moment first finishes, it clears the pool of the second as well.

aim/storage/types.py
```diff
@@ -66,6 +67,48 @@ class CustomObjectBase:
...
+
+
+class BLOB(Generic[T]):
+    def __init__(
+        self,
+        data: T = None,
+        loader: 'BLOBLoader' = None
+    ):
+        self.data: T = data
+        self.loader_f = loader
```
mahnerak, line 81: Let's have `loader_fn` and `transform_fn` for better readibility.

aim/sdk/uri_service.py
mihran113: For the first part: yeah, need to be changed. 
For the second part: because the URIs need to be transferred to the UI and back in the request body, that's why I sticked to strings.

aim/sdk/uri_service.py
mihran113: yeah, that will need an improvement in the future, but as of now, URI service is used by the UI only, so in that case we'll be fine, as each request creates it's own service.

aim/sdk/uri_service.py
alberttorosyan: @mahnerak we have a difference in results for the following two expressions:
`container.tree().subtree(resource_path).collect9)` and
`container.tree().collect(resource_path)`

lets say `resourse_path == a.b.c`
the 2nd one will return `{a: {b: {c: <VALUE>}}}`

alberttorosyan: for the above example
`container.tree().subtree('a').collect(('b', 'c'))` will return `<VALUE>`. I assume this comes from difference in implementation of PrefixView and RocksContainer

alberttorosyan (approved):
additional changes were pushed based on discussion with @mahnerak.
approving

PR #1002, opened 2021-11-11
Aim Images/Custom objects support

alberttorosyan:


mihran113 (changes requested):
Some changes to make:
- sdk/repo.py: use `Sequence` instead of `Metric` in `query_runs()` method to be able to iterate over all sequences of the run not only metrics
- sdk/utils.py: fix regex to include dots as well (add notes for future restriction of Custom Object's `AIM_NAME`s)

mihran113 (approved):
finally 🙏

Issue #999, opened 2021-11-11
Security vulnerabilty


haby0:
What would be the right contact to report a security vulnerabilty? thanks!

gorarakelyan:
@haby0 feel free to report any issues here.

SGevorg:
@haby0 how about you join our slack channel and let's discuss there ?

SGevorg:
@haby0  here is the link https://slack.aimstack.io/

Thanks for the contribution :).

gorarakelyan:
Resolved in Aim v3.1.0

PR #1003, opened 2021-11-12
Security issue fix for `/static-files/{path}` endpoint 

mihran113:
Fix security issue when incorrect path is given to the endpoint that serves static files which can lead to a leak of files (e.g. `/static-files/../../../../etc/passwd`)

### DIFF
--- a/aim/web/api/views.py
+++ b/aim/web/api/views.py
@@ -1,7 +1,9 @@
 import os
+from pathlib import Path
 
 from aim.web.api.utils import APIRouter  # wrapper for fastapi.APIRouter
 from fastapi.responses import FileResponse
+from fastapi import HTTPException
 
 statics_router = APIRouter()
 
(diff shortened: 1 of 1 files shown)


mihran113:
resolves #999

haby0:
If you use ```os.path.join(static_file_root, path)``` to join the path, the ```path``` parameter value cannot have ```/``` or ```\\```, Otherwise, the request will definitely make the result of ```if``` judgment as ```True```.

mihran113:
> If you use `os.path.join(static_file_root, path)` to join the path, the `path` parameter value cannot have `/` or `\\`, Otherwise, the request will definitely make the result of `if` judgment as `True`.

I think you meant that if the `path` starts with `/` or `\\` then `os.path.join` won't work as expected. Yeah, that's a valid concern, but our UI handles that. Anyways will change that with usual join to avoid any failures. Thanks for the comment.

alberttorosyan (approved):

Issue #861, opened 2021-10-16
Enforce tracked sequence to be homogeneous


alberttorosyan:
In Aim `Sequence` is defined as a series of homogeneous values (for example, `Metric` is a series of floating point values).
It is assumed that for the same sequence `Run.track()` is called for the objects of a same type. However this is not enforced.

Need to enhance the `track` method to check object's type before writing it to db. The following to be taken into account:
- tracking of a sequence values can be interrupted and continiued in a separate process
- empty list is a valid value for tracking

alberttorosyan:
The related code is not merged to `main` yet. Branch `feature/images-query-api` should be used meanwhile.

PR #1008, opened 2021-11-17
Ensure single sequence values are homogeneous

alberttorosyan:
Added object type checking in `Run.track()`. The logic is the following:
Per each run sequence there's a `SequenceInfo` in-memory object which holds values count and data type for the sequence.
When the first value is added, `SequenceInfo` is initialized. There are two possible cases:
1. The sequence is constructed from scratch, i.e. there's nothing written in sequence container yet. In this case data type is set to be equal to the first tracked value's type.
2. The sequence is continued on top of existing one. Here the data type is set to the one written in run's metadata container.  Since there might be repositories, with logged metrics generated prior to adding `dtype`, we treat missing value as if it was `dtype float`.

Values incompatible with the sequence data type are rejected and `ValueError` is raised.
There's a special handling when value type is list. In case when the first tracked value is an empty list, we cannot check inner object data type, and the sequence `dtype` is allowed to change. Once a non-empty list is passed to `Run.track()` sequence type is updated. No other object types are allowed since that, though empty lists are still valid input for that sequence.


alberttorosyan:
resolves #861

aim/sdk/run.py
```diff
@@ -371,17 +377,28 @@ def _track_impl(
...
-            max_idx = len(val_view)
-        step = step or max_idx
-
-        self.series_counters[metric.selector] = max_idx + 1
-
-        if max_idx == 0:
+        seq_info = self.sequence_info[metric.selector]
+        if not seq_info.initialized:
+            seq_info.count = len(val_view)
+            seq_info.trace_dtype = self.meta_run_tree.get(('traces', ctx.idx, name, 'dtype'), None)
```
mihran113, line 383: `sequence_dtype` instead of `trace_dtype`?

aim/sdk/run.py
```diff
@@ -371,17 +377,28 @@ def _track_impl(
...
+                self.meta_tree['traces_types', new_dtype, ctx.idx, name] = 1
+                seq_info.trace_dtype = self.meta_run_tree['traces', ctx.idx, name, 'dtype'] = new_dtype
+            compatible = check_types_compatibility(dtype, seq_info.trace_dtype, update_trace_dtype)
+            if not compatible:
+                raise ValueError(f'Cannot log value \'{value}\' on sequence \'{name}\'. Incompatible data types.')
+
+        step = step or seq_info.count
+
+        if seq_info.count == 0:
             # TODO [AT] check sequence is homogeneous & handle empty list case
```
mihran113, line 399: this `TODO` can be deleted, right?

aim/sdk/run.py
alberttorosyan: 👌

aim/sdk/run.py
alberttorosyan: yeah, that's the whole point of this PR

PR #1016, opened 2021-11-19
[feat] Additional properties for Aim Images built from PIL.Image.Image

alberttorosyan:
Added image mode and image type. Image type is hard-coded to 'PIL.Image'


mihran113 (approved):

gorarakelyan (approved):
lgtm

PR #1018, opened 2021-11-19
Improve run detail page header ux

VkoHov:
- Improve run detail header select dropdown
- Fixed table manage columns method
- Fix changing metrics page axis linear scale to log
- Add ability to search with Enter button in images explore
- Change resize mode to "resizable" after empty response


KaroMourad (approved):
lgtm

Issue #1025, opened 2021-11-22
Ability to cancel pending request


roubkar:
## 🚀 Feature

Provide an ability to cancel pending requests from UI.

### Motivation

The requests can sometimes take long to be fulfilled and there's no way to escape the pending state in UI.
Even worse case is when the latest query is being saved in the URL and you need to wait for the query result to return on page enter.

### Pitch

Would be great to have a button for this action. The `Search` button can be switched by a `Cancel` button while the request is in process.

PR #1026, opened 2021-11-22
Add ability to cancel pending request

roubkar:
- Replace the `Search` button with `Cancel` during a pending request.
- Send abort signal to the pending request on `Cancel` button click


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1025

rubenaprikyan (commented):
It seems we have code duplications?

roubkar:
> It seems we have code duplications?

Yeah, I've added the same functionality to as many places as we are using the search requests.
We need to go over app creation improvement and SelectForm component improvement which will also reduce duplications with other items related to them such as search request handling.

KaroMourad (approved):

VkoHov (approved):

PR #1029, opened 2021-11-22
Do not show warnings when readonly run container optimization fails

alberttorosyan:
Added silent flag to `exception_resistant` decorator


gorarakelyan (approved):
lgtm

PR #1032, opened 2021-11-23
Slice image sequence by index for the given steps range.

alberttorosyan:
Added method to select N values from the given [start, stop) range
of tracked steps. Entire range is loaded into memory so the number of
items can be identified and sliced correctly.


mihran113 (approved):
lgtm

PR #1039, opened 2021-11-23
Update images URI batch request delay

roubkar:
Changed images URIs get by batch request delay from `1s` to `300ms`


gorarakelyan (approved):
lgtm

PR #1040, opened 2021-11-23
Replace nan/inf/-inf values with None when returning metric batch.

alberttorosyan:


mihran113 (approved):

aim/web/api/runs/utils.py
```diff
@@ -195,6 +195,8 @@ def collect_requested_metric_traces(run: Run, requested_traces: List[TraceBase],
 
         iters, values = trace.values.sparse_list()
 
+        values = list(map(lambda x: x if float('-inf') < x < float('inf') and x == x else None, values))
```
mahnerak, line 198: Not a proper place to fix the bug with JSON error.

mahnerak: If you're getting a JSON serialization error while sending an response from backend using JSON - it's issue with json, not the storage. The SDK is designed to support Inf:
```py
>>> aim.storage.encoding.encode(inf)
b'\x03\x00\x00\x00\x00\x00\x00\xf0\x7f'
>>> aim.storage.encoding.decode(b'\x03\x00\x00\x00\x00\x00\x00\xf0\x7f')
inf
>>> aim.storage.encoding.encode(nan)
b'\x03\x00\x00\x00\x00\x00\x00\xf8\x7f'
>>> aim.storage.encoding.decode(b'\x03\x00\x00\x00\x00\x00\x00\xf8\x7f')
nan
```

mihran113: Yeah, @mahnerak we clearly know that, but at this point to not open an additional task for UI team we decided to handle it on the server side this way instead of switching to streaming response. You can have a look at a discussion in slack channel.

alberttorosyan: @mahnerak that's why the fix is not int he SDK, but in web API utils. There are two sides of the issue:
1. Current implementation of FastAPI JSON response forbids `nan/inf/-inf` values.
2. UI widgets do not handle those values. 
Possible solutions:
1. Replace `JSONReponse` with custom class allowing to serialize those values. This will require additional changes in UI.
2. Replace values with `None` (works fine in UI). This doesn't affect SDK usage.

alberttorosyan: CC @roubkar

alberttorosyan: I'm merging this now @mahnerak.

PR #1042, opened 2021-11-23
Support of Image input additional data sources.

alberttorosyan:
Added support for tf.Tensor and torch.Tensor.
Added support for numpy.array.
Refactored PIL.Image support.

### DIFF
--- a/aim/sdk/objects/image.py
+++ b/aim/sdk/objects/image.py
@@ -113,10 +113,14 @@ class Image(CustomObject):
         self.storage['width'], self.storage['height'] = pil_image.size
 
     def _from_numpy_array(self, array: np.ndarray):
-        shape = array.shape
-        if len(shape) not in {2, 3}:
-            raise ValueError('Cannot convert to aim.Image. array must have 2-D or 3-D shape.')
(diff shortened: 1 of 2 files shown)


aim/sdk/objects/image.py
```diff
@@ -102,9 +101,76 @@ def json(self):
...
-    # TODO support other sources as well
-    raise ValueError
+        self.storage['data'] = BLOB(data=image_bytes_p.getvalue())
+        self.storage['source'] = 'PIL.Image'
+        self.storage['mode'] = pil_image.mode
+        self.storage['format'] = 'png'
+        self.storage['width'], self.storage['height'] = pil_image.size
+
+    def _from_numpy_array(self, array: np.ndarray):
+        if array.ndim not in {2, 3}:
```
gorarakelyan, line 116: what about rgba mode? ndim=4
doesn't ndim of grayscale mode equal to 1?

alberttorosyan: @gorarakelyan `ndim` is still 3 in this case, the value of third dimension would be 4

gorarakelyan (approved):
lgtm

PR #1045, opened 2021-11-24
Server side fixes and additions

mihran113:
Provide `-w` or `--workers` option for `aim up` command to specify `uvicorn` workers count and make images/metrics search endpoints async to avoid memory piling up


alberttorosyan (approved):
lgtm

Issue #1041, opened 2021-11-23
Ignore outliers action is not working as expected


roubkar:
## 🐛 Bug

In `Metrics Explorer` the remove outliers feature for the line chart is ignoring the outliers in terms of Y-axis values for all the lines (the entire domain). Due to this entire lines are being ignored.

### To reproduce

- Go to `Metrics Explorer`
- Search for multiple metrics (can even be system metrics, e.g. Memory and GPU Temperature) so that their values/lines are polarized on the Y-axis scale
- Enable the `Ignore Outliers` button

### Expected behavior

Outliers should be ignored in a way that will still be able to see all the lines.

### Environment

- Aim Version 3.1.0

gorarakelyan:
Most probably outliers calculation is not working properly.
I got several lines totally ignored while they were not outliers.

PR #1047, opened 2021-11-24
Improve line chart outliers calculation

roubkar:
- Apply `remove outliers` functionality on min and max values of all lines
- Improve X-axis ticks range calculation for time alignment


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1041

gorarakelyan (approved):
lgtm

PR #1049, opened 2021-11-24
Performance tests fixes

mihran113:
Restructure performance tests to use in-place logs instead of downloading from S3 and extracting and adjust baselines for newly added tests. Also added env variable to fallback to the previous download/extract approach.


alberttorosyan (approved):
lgtm

PR #1061, opened 2021-11-26
Feature/auto resource cleanup

mihran113:


aim/ext/cleanup/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        ...
+
+    @staticmethod
+    def _cleanup():
+        """
+        Actually cleanup all the remaining objects in order of specified
+        priorities. This is called automatically by atexit.
+        It is not meant tobe called manually.
+        """
+        logger.debug("Cleaning up...  Found {len(finalizers)} finalizers")
```
mahnerak, line 72: Missed `f'...`. Please fix the quotes.

aim/sdk/repo.py
```diff
@@ -81,16 +108,18 @@ def __init__(self, path: str, *, read_only: bool = None, init: bool = False):
                 unmount_remote_repo(self.root_path, self._mount_root)
             raise RuntimeError(f'Cannot find repository \'{self.path}\'. Please init first.')
 
-        self.container_pool: Dict[ContainerConfig, Container] = WeakValueDictionary()
+        self.container_pool: WeakValueDictionary[ContainerConfig, Container] = WeakValueDictionary()
         self.persistent_pool: Dict[ContainerConfig, Container] = dict()
-        self.container_view_pool: Dict[ContainerConfig, Container] = WeakValueDictionary()
+        self.container_view_pool: WeakValueDictionary[ContainerConfig, Container] = WeakValueDictionary()
```
mahnerak, line 113: I remember having issues with genertic typings of `WeakValueDictionary`.
Please check these with Python3.6.

mahnerak: Yup, tests do not pass with Python 3.6. We can either use Dict or a more generic MutableMapping.

aim/ext/cleanup/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    def join(self):
+        while True:
+            try:
+                # Pressing Ctrl-C will raise KeyboardInterrupt and stop only
+                # the `Thread.join()` call. The target function of `RobustExec`
+                # will continue to run.
+                return super().join()
+            except KeyboardInterrupt:
+                logger.warning("Received Ctrl-C. Closing gracefully.")
```
mahnerak, line 29: @gorarakelyan This fully blocks force-Ctrl-C-aborting.
We can use this implementation now, but we should think of a way of letting user to abort the running process immediately.

aim/sdk/repo.py
mihran113: Tests do not pass, but somewhere else, not here. `WeakKeyDictionary` is the problematic one.

mahnerak: Yes, there was an issue when using `WeakKeyDictionary[]` typing with generic `T`. I believe we should avoid using `WeakValueDictionary` and `WeakKeyDictionary` as type annotations (including here).

mihran113: yeah, will change back to `Dict`

aim/ext/cleanup/__init__.py
mihran113: done ✅

aim/ext/cleanup/__init__.py
gorarakelyan: What about displaying prompt message asking whether user wants to terminate the process immediately?

alberttorosyan (approved):

PR #1062, opened 2021-11-29
Fix images explorer tooltip hiding functionality

KaroMourad:
Improved ChartPopover component

Added mousemove function on images container and on images.

- Image mousemove with stopPropagation will call for tooltip positioning.
- Images container mousemove will call every time(on hovering) if the mouse position is not on image , because of stopPropagation, which will prevent calling the container mousemove and will close tooltip in container mousemove function


roubkar (approved):

VkoHov (approved):

VkoHov:
lgtm

PR #1085, opened 2021-12-02
Added a support for tracking pythonic dictionary objects

osoblanco:
Added a custom Dictionary object to track throughout runs.


gorarakelyan:
Closing the PR as the code depends on the legacy SDK.

Issue #1070, opened 2021-12-01
Ability to track plotly plot objects


alberttorosyan:
Need to define aim `CustomObject` which will take an arbitrary `plotly` plot and serialize its json dump to aim storage.
Support for querying tracked sequences, and web APIs to retrieve queried data.

PR #1087, opened 2021-12-03
1070 | Support Plotly figure objects

devfox-se:
implements #1070


aim/sdk/objects/plotly.py
```diff
@@ -0,0 +1,33 @@
+from aim.sdk.num_utils import inst_has_typename
+from aim.storage.object import CustomObject
+from aim.storage.types import BLOB
+
+
+@CustomObject.alias("aim.plotly")
+class Plotly(CustomObject):
+    AIM_NAME = "aim.plotly"
```
alberttorosyan, line 8: lets rename this to `aim.figure` or similar
in the future we might support more inputs (different from `plotly`)

aim/sdk/objects/plotly.py
```diff
@@ -0,0 +1,33 @@
...
+    def _prepare(self, obj):
+        assert hasattr(obj, "to_json")
+
+        self.storage["data"] = BLOB(data=obj.to_json())
+
+    def json(self):
+        blob_data = self.storage["data"]
+        return blob_data.data
+
+    def to_figure(self):
```
alberttorosyan, line 27: to_plotly_figure? 🤔

aim/sdk/objects/plotly.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if inst_has_typename(obj, ["Figure", "BaseFigure"]):
+            self._prepare(obj)
+        else:
+            raise TypeError("Object is not a Plotly instance")
+
+    def _prepare(self, obj):
+        assert hasattr(obj, "to_json")
+
+        self.storage["data"] = BLOB(data=obj.to_json())
```
alberttorosyan, line 21: please add the `source='plotly'` and `format='raw_json' to storage to make this future-proof.once we change format and/or add new types of input, this will be required.

aim/sdk/plotly_sequence.py
```diff
@@ -0,0 +1,38 @@
+from typing import Union, Tuple
+
+from aim.sdk.sequence import Sequence
+from aim.sdk.objects.plotly import Plotly
+
+
+class Plotlys(Sequence):
```
alberttorosyan, line 7: Plotlys  -> Figures

aim/sdk/plotly_sequence.py
```diff
@@ -0,0 +1,38 @@
...
+    """Class representing series of Plotly objects or Plotly lists."""
+
+    @classmethod
+    def allowed_dtypes(cls) -> Union[str, Tuple[str, ...]]:
+        plotly_typename = Plotly.get_typename()
+        return plotly_typename, f'list({plotly_typename})'
```
alberttorosyan, line 13: @gorarakelyan @mahnerak should we support lists of plotly figures as an input to `Run.track`?

alberttorosyan (changes requested):
general feedback: please use term figure/Figure for all aim classes and APIs. This will allow us to add new input sources (similar to plotly figures) without adding/changing the custom object

aim/sdk/plotly_sequence.py
devfox-se: yeah seems more logical - i was struggling to come up with a name 🤣

aim/sdk/plotly_sequence.py
gorarakelyan: I do not think so. If user wants to track several plotly figures it will be better to log them in separate sequences imo.
In case of audios/videos list of values is supported because usually users track batches of samples not a single sample.
@alberttorosyan @devfox-se ^

alberttorosyan: please remove `f'list({plotly_typename})'` from supported types based on @gorarakelyan's feedback.

aim/sdk/plotly_sequence.py
```diff
@@ -0,0 +1,38 @@
...
+        return self._meta_tree['first_step']
+
+    def last_step(self):
+        """Get sequence tracked last step.
+
+        Required to implement ranged and sliced data fetching.
+        """
+        return self._meta_tree['last_step']
+
+    def record_length(self):
```
alberttorosyan, line 33: we don't need this method since only single figures are allowed to track

alberttorosyan (commented):
left a couple of comments. the rest looks good

aim/sdk/plotly_sequence.py
devfox-se: removed with bunch of refactoring that it has caused

alberttorosyan (approved):
looks good!

alberttorosyan (approved):

PR #1090, opened 2021-12-03
Fix boolean encoding

mahnerak:
Fixes the issue with encoding of booleans always converting to integer.

Previously, even if we decode the encoded buffer of `0.0` to `True`, detecting inequality with `assertEqual` will not work.
This PR adds more type-strict tests to avoid this kind of issues in the future.


alberttorosyan (approved):
lgtm

rubenaprikyan (approved):
lgtm!

mihran113 (approved):

PR #1102, opened 2021-12-07
Add distributions tracking basics

gorarakelyan:


alberttorosyan (approved):
lgtm.
@gorarakelyan is it worth to add a comment that the argument to the `Distribution` constructor should be an iterable of scalars?

gorarakelyan:
@alberttorosyan I guess yes. Will add the note, thanks

SGevorg (approved):
lgtm.

PR #1106, opened 2021-12-08
Add `tf.tensor` support for tracking

mihran113:


mihran113:
Resolves #1107

gorarakelyan (approved):
looks great

PR #1119, opened 2021-12-10
Sort context names alphabetically

rubenaprikyan:
- Show contexts list ordered by stringified value alphabetically

### DIFF
--- a/aim/web/ui/src/services/models/runs/util.ts
+++ b/aim/web/ui/src/services/models/runs/util.ts
@@ -34,6 +34,9 @@ export function getMenuItemFromRawInfo(
   function sortOrder(a: IMenuItem, b: IMenuItem) {
     const name1 = a.name.toUpperCase();
     const name2 = b.name.toUpperCase();
+    if (name2 === 'EMPTY CONTEXT') {
+      return 0;
+    }
     if (name1 < name2) {
       return -1;
     }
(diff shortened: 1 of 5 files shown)


Hamik25 (approved):
Looks good!

arsengit (approved):

aim/web/ui/src/services/models/runs/util.ts
```diff
@@ -47,6 +48,16 @@ export function getMenuItemFromRawInfo(
 
   const data: IMenuItem[] = [];
 
+  const sortOrder = alphabeticalSortComparator({
+    orderBy: 'name',
+    additionalCompare: (name1: string, name2: string) => {
+      if (name2 === 'EMPTY CONTEXT') {
+        return 0;
```
roubkar, line 55: 0 or -1 ?

rubenaprikyan:
Resolves #1122

aim/web/ui/src/services/models/runs/util.ts
rubenaprikyan: It need to show at the beginning of the list.
In the menu list of TraceVisualizer if there is an empty context, it is always positioned at the beginning of the list, so it need to keep position the same.

Please also see the comment in line 108, 109 of this file

roubkar (approved):

Issue #1122, opened 2021-12-10
Show contexts listed in alphabetical order


rubenaprikyan:
## 🚀 Feature

Show contexts listed in alphabetical order

### Motivation
It would be convenient to see alphabetically ordered  lists for the values in Metric, Params, Images dropdown's in the  corresponding explorers

![Screen Shot 2021-12-10 at 11 47 44 PM](https://user-images.githubusercontent.com/21033329/145633528-37433861-7e0d-4067-b527-4a482e75d96a.png)
ing explorers.

![Screen Shot 2021-12-10 at 11 47 55 PM](https://user-images.githubusercontent.com/21033329/145633381-fb9e4a78-bf72-447c-8ea0-e24ebec8bde4.png)
20-a707683071e1.png

### Pitch
Show contexts, params, metrics lists in alphabetical order

mahnerak:
🙏

PR #1123, opened 2021-12-10
Sort alphabetically ordered lists for the values in Metric, Params, Images dropdowns

rubenaprikyan:
- Sort alphabetically ordered lists for the values in Metric, Params, Images dropdown's
- Change `No context` texts to `Empty context`
- Show system metric as items of group System


rubenaprikyan:
Resolves #1122

roubkar:
I think all the `metrics` should come before `params` in this list.

Screen Shot 2021-12-12 at 18 17 03

---

Also, let's add sorting based on the search string occurrence index.
We have similar functionality in grouping popover.

roubkar (approved):

PR #1127, opened 2021-12-12
Aim File/Audio tracking

devfox-se:


aim/sdk/objects/audio.py
```diff
@@ -0,0 +1,96 @@
...
+        self._prepare(data, caption=caption, format=audio_format)
+
+    def _prepare(self, buffer, **extra):
+        assert isinstance(buffer, io.BytesIO)
+
+        for k, v in extra.items():
+            self.storage[k] = v
+        self.storage['source'] = 'audio'
+        self.storage['bytes'] = BLOB(data=buffer.read())
+        self.storage['size'] = self.bytes_to_readable(buffer.getbuffer().nbytes)
```
alberttorosyan, line 61: why do we need to store 'size'? 'duration' would probably make sense, but I guess we cannot retrieve that easily.

aim/sdk/objects/audio.py
```diff
@@ -0,0 +1,96 @@
...
+class Audio(CustomObject):
+    AIM_NAME = 'aim.audio'
+
+    # supported audio formats
+    UNKNOWN = 0
+    MP3 = 1
+    WAV = 2
+    FLAC = 3
+
+    __audio_format_map = {
```
alberttorosyan, line 22: let's use strings as inputs. we still have to do validation of formats, but users will prefer not to learn new naming, enums, etc.
besides, the format we'll store and later read from storage will be string.

aim/sdk/objects/audio.py
```diff
@@ -0,0 +1,96 @@
...
+                logger.warning(f'Parameter "rate" is not provided! Using default: {rate}')
+            bs = wavfile.write(rate, data)
+            data = bs
+        else:
+            # act as a regular file with enforced audio format definition by user side
+            audio_format = self.__audio_format_map.get(audio_format)
+            if not audio_format:
+                raise ValueError('Invalid audio format is provided.')
+
+        if not isinstance(data, io.BytesIO):
```
alberttorosyan, line 49: we need to support raw `bytes` and file paths as well.

aim/sdk/objects/audio.py
```diff
@@ -0,0 +1,96 @@
...
+            raise TypeError('Content is not a byte-stream object')
+
+        self._prepare(data, caption=caption, format=audio_format)
+
+    def _prepare(self, buffer, **extra):
+        assert isinstance(buffer, io.BytesIO)
+
+        for k, v in extra.items():
+            self.storage[k] = v
+        self.storage['source'] = 'audio'
```
alberttorosyan, line 59: I guess this is redundant. We already storing similar info as object typename.
What would be more important is to store the actual input source. I.e.
```
self.storage['source'] = 'numpy'
self.storage['source'] = 'file'
```
and based on each source, provide additional metadata, such as file path in case of 'file'.

aim/sdk/objects/audio.py
```diff
@@ -0,0 +1,96 @@
...
+        self.storage['bytes'] = BLOB(data=buffer.read())
```
alberttorosyan, line 60: please rename to `'data'` to be aligned with other media objects

aim/sdk/objects/audio.py
```diff
@@ -0,0 +1,96 @@
...
+            num /= 1024.0
+        return f'{num:.1f}Y{suffix}'
+
+    def get(self) -> io.BytesIO:
+        bs = self.storage.get('bytes')
+        if not bs:
+            return io.BytesIO()
+        return io.BytesIO(bytes(bs))
+
+    def dump(self, path):
```
alberttorosyan, line 89: let's not add this interface yet.

aim/web/api/runs/audio_utils.py
```diff
@@ -0,0 +1,193 @@
...
+def sliced_audio_record(values: Iterable[Audio], _slice: slice) -> Iterable[Audio]:
+    yield from zip(range(_slice.start, _slice.stop, _slice.step), values[_slice])
+
+
+def audio_record_to_encodable(obj, trace, step, index=0):
+    resource_path = generate_resource_path(trace.values.tree.container, (step, 'data'))
+    return [{
+        'index': index,
+        'format': obj.storage.get('format'),
+        'size': obj.storage.get('size'),
```
alberttorosyan, line 29: wrote a comment above. do we need `'size'`?

alberttorosyan (changes requested):
@devfox-se Arthur jan, left some comments.

aim/web/api/runs/audio_utils.py
devfox-se: Nice addition but completely optional

aim/sdk/objects/audio.py
devfox-se: removed size - wont add duration

aim/sdk/objects/audio.py
devfox-se: I dont think this matter because everything is converted to bytestream

aim/sdk/objects/audio.py
```diff
@@ -14,36 +14,32 @@ class Audio(CustomObject):
...
         rate = kwargs.get('rate', 22050)
-        audio_format = kwargs.get('format', self.UNKNOWN)
+        audio_format = kwargs.get('format', self.UNKNOWN).lower()
 
         if inst_has_typename(data, ['ndarray.numpy']):
             # Currently, only WAV audio formats are supported for numpy
-            audio_format = self.__audio_format_map[self.WAV]
+            audio_format = self.WAV
             if 'rate' not in kwargs:
                 logger.warning(f'Parameter "rate" is not provided! Using default: {rate}')
```
alberttorosyan, line 36: maybe level info? 🤔 I guess this can happen quite frequently.

aim/sdk/objects/audio.py
```diff
@@ -14,36 +14,32 @@ class Audio(CustomObject):
...
             bs = wavfile.write(rate, data)
             data = bs
         else:
             # act as a regular file with enforced audio format definition by user side
-            audio_format = self.__audio_format_map.get(audio_format)
             if not audio_format:
+                raise ValueError('Audio format must be provided.')
+            if audio_format not in self.audio_formats:
                 raise ValueError('Invalid audio format is provided.')
```
alberttorosyan: Invalid audio format is provided. Must be one of 'WAV', 'MP3', 'FLAC'

aim/sdk/objects/audio.py
alberttorosyan: ok, lets remove this key then

aim/sdk/objects/io/wavfile.py
```diff
@@ -0,0 +1,715 @@
+"""
+NOTICE: All credits to author of SciPy package
```
alberttorosyan, line 2: @gorarakelyan here we use the code from scipy package. It's distributed under GPL license.
https://github.com/scipy/scipy/blob/master/LICENSE.txt
Could you please confirm that the copyright notice here is OK? If not, what should be added/changed.

alberttorosyan (changes requested):
looks good! commented for final small improvements

aim/sdk/objects/io/wavfile.py
gorarakelyan: I guess it is ok. 

>The Free Software Foundation considers the Apache License, Version 2.0 to be a free software license, compatible with version 3 of the GPL.

Source: https://www.apache.org/licenses/GPL-compatibility.html

alberttorosyan (approved):
lgtm

PR #1133, opened 2021-12-14
Added focused/hovered states style of search-inputs

KaroMourad:
When hovering over on table rows or chart, chart tooltip opens and Material UI Popover component changes focus on tooltip

Fix: 
- disable auto-focus on chart tooltip
- tuned Dropdown Component options style

Tested on browsers:  Chrome, Safari , Firefox


VkoHov:
lgtm

PR #1152, opened 2021-12-17
Fix getValue function to show correct chart title

KaroMourad:
Fix:
- getValue function to show correct chart title
- formatValue function to show correct formatted data


Hamik25 (approved):

gorarakelyan (approved):
lgtm

PR #1154, opened 2021-12-19
Fix issue with not providing point density value in live-update query

rubenaprikyan:
- Fix issue with not providing point density value in live-update query


rubenaprikyan:
Resolves #1155

roubkar (approved):
lgtm

PR #1164, opened 2021-12-22
Fix issue with showing range panel in Images Explorer

roubkar:


aim/web/ui/src/services/models/imagesExplore/imagesExploreAppModel.ts
```diff
@@ -1965,7 +1960,7 @@ const imagesExploreAppModel = {
   onImageSizeChange,
   onImageRenderingChange,
   onImageAlignmentChange,
-  isRangePanelShow,
+  showRangePanel,
   getGroupingSelectOptions,
```
rubenaprikyan, line 1964: looks good!

rubenaprikyan (approved):
Looks good!

PR #1169, opened 2021-12-24
[feat] image object enhancements

devfox-se:
Adds features:

- Read image content from file path
- Let user define extra PIL save parameters

### DIFF
--- a/aim/sdk/objects/image.py
+++ b/aim/sdk/objects/image.py
@@ -140,7 +140,7 @@ class Image(CustomObject):
                 raise
             elif not Image.FLAG_WARN_RGBA_RGB:
                 logger.warning(f'Failed to save the image due to the following error: {exc}')
-                logger.warning('Attempting to convert RGBA -> RGB')
+                logger.warning(f'Attempting to convert mode "{pil_image.mode}" to "RGB"')
                 Image.FLAG_WARN_RGBA_RGB = True
 
             alpha = pil_image.convert('RGBA').split()[-1]  # Get only alpha
--- a/aim/sdk/objects/image.py
+++ b/aim/sdk/objects/image.py
@@ -1,3 +1,5 @@
+import os.path
+
 from PIL import Image as PILImage
 
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


aim/sdk/objects/image.py
```diff
@@ -101,29 +106,52 @@ def json(self):
...
+                # Remove transparency
+                alpha = pil_image.convert('RGBA').split()[-1]  # Get only alpha
+                background = PILImage.new('RGBA', pil_image.size, (255, 255, 255, 255))
+                background.paste(pil_image, mask=alpha)
+                pil_image = background.convert('RGB')
+
+            params = {
+                'format': 'jpeg',
+                'optimize': True,
+                'quality': 80
```
alberttorosyan, line 124: @devfox-se how much compression we get with this values?

aim/sdk/objects/image.py
```diff
@@ -101,29 +106,52 @@ def json(self):
...
 
-    def _from_pil_image(self, pil_image: PILImage.Image):
+    def _from_pil_image(self, pil_image: PILImage.Image, compress=False):
         assert isinstance(pil_image, PILImage.Image)
-        image_bytes_p = BytesIO()
-        pil_image.save(image_bytes_p, format='png')
+        img_container = BytesIO()
+
+        if compress:
+            if pil_image.mode == 'RGBA' or isinstance(pil_image.info.get('transparency'), bytes):
```
alberttorosyan, line 114: @devfox-se please comment why we need this step

aim/sdk/objects/image.py
```diff
@@ -101,29 +106,52 @@ def json(self):
             'height': self.height,
         }
 
-    def _from_pil_image(self, pil_image: PILImage.Image):
+    def _from_pil_image(self, pil_image: PILImage.Image, compress=False):
```
alberttorosyan, line 109: since this and the rest of `_from_*` methods are called from `__init__` where we explicitly pass compress, no need for default value.

aim/sdk/objects/image.py
devfox-se: @alberttorosyan already have a comment underneath that line:)

aim/sdk/objects/image.py
devfox-se: On the second thought it's better to let the user define format and the quality

aim/sdk/objects/image.py
```diff
@@ -9,29 +12,46 @@
...
 
 @CustomObject.alias('aim.image')
 class Image(CustomObject):
-    """Image object used to store image objects in Aim repository..
+    """Image object used to store image objects in Aim repository...
 
     Args:
          image (:obj:): pillow `Image` object or `torch.Tensor` or `numpy.array` used to construct `aim.Image`.
          caption (:obj:`str`, optional): Optional image caption. '' by default.
+         format, quality, optimize: PIL image save params. For more information refer to PIL documentation.
```
alberttorosyan, line 25: please also list the default values here

alberttorosyan (approved):
left a small comment re: docstrings

the rest looks great

alberttorosyan:
resolves #1233

Issue #1175, opened 2022-01-03
Aim adapter for pytorch lightning fails when trying to log `omegaconf`.`DictConfig`


alberttorosyan:
This issue is a part of #404 and aims to resolve errors when pytorch lightning adapter is used to log hydra configs with aim.
Detailed description and error stack trace can be found [here](https://github.com/aimhubio/aim/issues/404#issuecomment-1003913559)

In order to resolve the issue `aim/sdk/adapters/pytorch_lightning.py` should check the type of logged params end convert to dict if needed.

Issue #404, opened 2021-03-25
Add support for hydra configs


gorarakelyan:
It would be great to support logging hydra(https://github.com/facebookresearch/hydra) configs directly without preliminary conversion to python dicts. For example `aim.set_params(hydra_cfg)`.

felixkreuk:
Would the `set_params` method of `Session` class be a good spot for this?
Thinking about it, it would introduce an OmegaConf dependency (hydra uses dictConfig), is that something you want?

gorarakelyan:
Yea, probably OmegaConf is the way to go. 

re `set_params`: For logging configs set_params is used currently, do you think there is a better way/interface of integration with hydra?

Thanks for sharing details. Really appreciate your help 🙏

haritsahm:
I got this error when using pytorch-lightning+hydra and aim. So aim will log the params but DictConfig from OmegaConf is not supported. How do I resolve this?

```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 252, in run_and_report
    assert mdl is not None
AssertionError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "run.py", line 34, in <module>
    main()
  File "/opt/conda/lib/python3.8/site-packages/hydra/main.py", line 48, in decorated_main
    _run_hydra(
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 377, in _run_hydra
    run_and_report(
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 294, in run_and_report
    raise ex
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 211, in run_and_report
    return func()
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 378, in <lambda>
    lambda: hydra.run(
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/hydra.py", line 111, in run
    _ = ret.return_value
  File "/opt/conda/lib/python3.8/site-packages/hydra/core/utils.py", line 233, in return_value
    raise self._return_value
  File "/opt/conda/lib/python3.8/site-packages/hydra/core/utils.py", line 160, in run_job
    ret.return_value = task_function(task_cfg)
  File "run.py", line 30, in main
    return train(config)
[...]

PR #1177, opened 2022-01-04
(fix) handling omegaconf params

devfox-se:
fixes bug as mentioned in #404


alberttorosyan (approved):
lgtm

lets also think about supporting other types of configs (basically anything which has dict protocol)

alberttorosyan:
resolves #1175

Issue #1170, opened 2021-12-28
[UI][feature] ability to remove runs


mglowacki100:
## 🚀 Feature
In "Runs" for every row (run) there could be trash icon to remove run from db.

### Motivation
- easy removal of failed experiments 
- stop-gap solution for distributed setting for which I didn't figure out yet, how to log results only from single GPU instead of four (it clutters interface)

### Alternative
- provide info how to do it directly on db

gorarakelyan:
hey @mglowacki100, development of the feature of hard-deleting runs via SDK is already in progress. Once it is done, will definitely add an option of deleting runs from UI as well. Thanks for the feedback 🙏

gloryVine:
@gorarakelyan Imo, this should have really high priority. I have been 'using' aim for a few months now, but essentially I end up wiping my database daily because its cluttered by runs which for some reason or another have no value to me: Sometimes it is because I had an error in the code which I only catched after the run was already over, at other times it's just because some runs were more informative than others, and I want to only focus on the latter ones. Since the aim interface is super cluttered by the uninformative runs, I copy the informative runs' values on paper and wipe the database. I think this is defeating the entire purpose of using an experiment tracking tool, but I don't know what else to do. Can you please at least explain how to delete runs in the database for now.

alberttorosyan:
@gloryVine is the disk space an issue for you? If no and the problem is that you don't want to see some runs in the UI, there's a soft-delete functionality in aim. It is available from Run Details page "Settings" tab.

gloryVine:
Hey @alberttorosyan. No, disk space is not an issue for me at the moment. Could you tell me in more detail where to look for the soft-delete functionality. By 'Run Details' page do you mean the 'Runs' tab which by default is on the left? If so, I don't see a 'Settings' tab there, even though I have the latest version of aim.

SGevorg:
Hi @gloryVine pls check out this [link](http://play.aimstack.io:10001/runs/7eac8cf) of a single run page on the demo site.
It has a settings tab where you can archive the run.
[...]

Issue #441, opened 2021-04-14
Delete experiments from disk through the dashboard


gorarakelyan:
Feature requested by Mohamed.

There is only an ability to archive experiments. It would be great to add an option of hard deleting experiments right from the UI.

PR #1184, opened 2022-01-05
[feat] Run hard-delete functionality

alberttorosyan:
Added `Repo` methods to delete Run data by given hash (list of hashes). The following data is removed:
- sqlite `Run` table entity for given hash
- run data from index db (chunks, BLOBs)
- meta/seqs containers data (including locks, progress files)

Added CLI and web API


alberttorosyan:
resolves #441
resolves #1170

aim/web/api/runs/views.py
```diff
@@ -481,3 +482,34 @@ async def remove_run_tag_api(run_id: str, tag_id: str, factory=Depends(object_fa
...
+        'status': 'OK'
+    }
+
+
+@runs_router.post('/delete-batch/')
+async def delete_runs_batch_api(runs_batch: RunsBatchIn):
+    # Get project
+    project = Project()
+    if not project.exists():
+        raise HTTPException(status_code=404)
```
devfox-se, line 507: we use this part in so many places maybe its time to wrap int under a decorator?

aim/cli/cli.py
```diff
@@ -1,12 +1,13 @@
 import click
 from click import core
 
-from aim.cli.configs import INIT_NAME, VERSION_NAME, UP_NAME, UPGRADE_NAME, REINDEX_NAME
+from aim.cli.configs import *  # noqa F403
```
devfox-se, line 4: I think it's better to do an explicit import like it was before

devfox-se (approved):
looks good - left some comments

aim/web/api/runs/views.py
alberttorosyan: agree @devfox-se. I do have this fixed in a different branch (for API unification).

aim/cli/runs/commands.py
```diff
@@ -0,0 +1,55 @@
...
+        exit(1)
+    chunks_dir = os.path.join(repo_path, '.aim', 'meta', 'chunks')
+    run_hashes = os.listdir(chunks_dir)
+
+    click.echo('\t'.join(run_hashes))
+    click.echo(f'Total {len(run_hashes)} runs.')
+
+
+@runs.command(name='rm')
+@click.argument('run_hash', nargs=-1, type=str)
```
gorarakelyan, line 38: `run_hash` or just `hash`? 🤔

alberttorosyan: @gorarakelyan will change.
actually this is shown just in a help message. the argument is positional so no need to specify `--run-hash`/`--hash`

alberttorosyan: UPD: @gorarakelyan cannot change, since the argument name should match the underlying command variable name. in that case `hash` will shadow python's built-in `hash()`.

aim/sdk/repo.py
```diff
@@ -337,6 +337,46 @@ def query_runs(self, query: str = '', paginated: bool = False, offset: str = Non
...
+            run_hash (:obj:`str`): Run to be deleted.
+
+        Returns:
+            True if run deleted successfully, False otherwise.
+        """
+        try:
+            self._delete_run(run_hash)
+            return True
+        except Exception:
+            return False
```
mahnerak, line 356: If we ignore any type of exceptions please at least log the error.

aim/sdk/repo.py
```diff
@@ -515,6 +555,24 @@ def _prepare_runs_cache(self):
...
         self.run_props_cache_hint = cache_name
 
+    def _delete_run(self, run_hash):
+        with self.structured_db:  # rollback db entity delete if subsequent actions fail.
+            # remove database entry
+            self.structured_db.delete_run(run_hash)
+
+            # remove data from index container
+            index_tree = self._get_index_container('meta', timeout=0).tree()
+            del index_tree.subtree(('meta', 'chunks'))[run_hash]
```
mahnerak, line 565: `del tree[key]` is designed to remove the BLOBs as well.

mahnerak (changes requested):
We need to change the way we deal with BLOBs.

aim/sdk/repo.py
alberttorosyan: valid point, thanks!

mahnerak (approved):
LGTM

PR #1187, opened 2022-01-09
[fix] Run on-the-fly WAL flushes and other optimizations for union DB.

alberttorosyan:
While the `RocksContainer` should run above-mentioned optimizations when db is accessed, this was not the case due to inheritance. This fix makes sure that WAL flush is run regardless how the `RocksContainer` is used; as a separate object or a part of union DB.


mahnerak (approved):
LGTM

PR #1193, opened 2022-01-13
[feat] TF Event log handling

devfox-se:


aim/cli/up/commands.py
```diff
@@ -115,6 +118,76 @@ def up(dev, host, port, workers, repo, tf_logs, ssl_keyfile, ssl_certfile, base_
     click.echo('Press Ctrl+C to exit')
 
     try:
+        if tf_logs:
```
alberttorosyan, line 121: lets remove the `.json` suffix

aim/cli/up/commands.py
```diff
@@ -115,6 +118,76 @@ def up(dev, host, port, workers, repo, tf_logs, ssl_keyfile, ssl_certfile, base_
...
+
+                    if hashsum != run_tf_hashsums.get(event, ''):
+                        # Something has changed or hasn't been processed before
+                        events_to_process.append(event)
+                    run_tf_hashsums[event] = hashsum
+                run['tf_hashsums'] = run_tf_hashsums
+
+                if events_to_process:
+                    # This import statement takes long to complete
+                    import tensorflow as tf
```
alberttorosyan, line 167: please wrap in try-except block since tensorflow is not a required dependency

aim/cli/up/commands.py
```diff
@@ -115,6 +118,76 @@ def up(dev, host, port, workers, repo, tf_logs, ssl_keyfile, ssl_certfile, base_
...
+                    from tensorflow.python.framework import tensor_util
+                    from tensorflow.python.summary.summary_iterator import summary_iterator
+
+                    for event in events_to_process:
```
alberttorosyan, line 171: lets change a naming a bit
event -> events_file
e -> event
v -> value

aim/cli/up/commands.py
```diff
@@ -115,6 +118,76 @@ def up(dev, host, port, workers, repo, tf_logs, ssl_keyfile, ssl_certfile, base_
...
+                        for e in summary_iterator(event):
+                            timestamp = e.wall_time
+                            step = e.step
+                            for v in e.summary.value:
+                                tag = v.tag
+                                plugin_name = v.metadata.plugin_data.plugin_name
+                                if plugin_name == 'images':
```
alberttorosyan, line 178: can we check the `tensor` shape and if it's a 4-D tensor, treat it as an image list?

aim/cli/up/commands.py
```diff
@@ -115,6 +118,76 @@ def up(dev, host, port, workers, repo, tf_logs, ssl_keyfile, ssl_certfile, base_
...
+                for event in events:
+                    # Calculate file hash
+                    chunk_size = 8192
+                    file_hash = hashlib.md5()
+                    with open(event, 'rb') as FS:
+                        chunk = FS.read(chunk_size)
+                        while chunk:
+                            file_hash.update(chunk)
+                            chunk = FS.read(chunk_size)
+                    hashsum = file_hash.hexdigest()
```
alberttorosyan, line 157: question: what is the case when event file has been changed after tf logs are converted?

alberttorosyan: I think we can keep things simple here. Just keep the most recent conversion time (per run) and process files modified after that.

alberttorosyan (changes requested):
@devfox-se jan, left some comments.

aim/cli/up/commands.py
devfox-se: I didnt get this

aim/cli/up/commands.py
devfox-se: fixed duplicate values in 66c5229

aim/cli/up/utils.py
```diff
@@ -27,3 +33,118 @@ def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile):
...
+            os.path.join(root, file)
+            for root, dirs, files in os.walk(path)
+            for file in files if file.endswith('v2')
+        ]
+
+        for event in events:
+            # Calculate file hash
+            chunk_size = 8192
+            file_hash = hashlib.md5()
+            with open(event, 'rb') as FS:
```
mahnerak, line 97: Tensorboard logs can contain gigabytes of data. Not sure if it's good idea to calculate md5 each time.

aim/cli/up/commands.py
devfox-se: Just figured out the kernel limitation only applies to the creation time so we're good to stick with modification time, will replace with hashsum generation part

aim/cli/up/utils.py
devfox-se: thanks for the input - I have replaced the logic to use the file modification time

aim/cli/up/commands.py
devfox-se: Cannot seem to resolve this,

aim/cli/up/utils.py
```diff
@@ -27,3 +32,112 @@ def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile):
...
+        tf_logs_cache = {}
+
+    for run_path in os.listdir(tf_logs):
+        path = os.path.join(tf_logs, run_path)
+        if path not in tf_logs_cache:
+            tf_logs_cache[path] = {}
+
+        run_cache = tf_logs_cache[path]
+        if run_cache:
+            run = Run(repo=repo_inst, run_hash=run_cache['run_hash'])
```
alberttorosyan, line 75: need to pass `system_tracking_interval=None` to disable resource usage tracking.

aim/cli/up/utils.py
```diff
@@ -27,3 +32,112 @@ def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile):
...
+        else:
+            run = Run(repo=repo_inst)
```
alberttorosyan, line 77: same here

aim/cli/up/utils.py
```diff
@@ -27,3 +32,112 @@ def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile):
...
+                    elif plugin_name == 'scalars':
+                        track_val = value.simple_value
+                    else:
+                        # unsupported plugin type
+                        continue
+                    run_tf_event['values'][value_id] = {
+                        'step': step,
+                        'timestamp': timestamp
+                    }
+                    run.track(track_val, name=tag, step=step)
```
alberttorosyan, line 139: lets use `run._track_impl()` in order to pass `event.wall_time`. otherwise we will log the time roughly about `aim up --tf_logs` was ran. 
the similar case is when we convert aim v2 logs (https://github.com/aimhubio/aim/blob/9ba5fdca36e69ab6cbb74296333c2f8bc56392a4/aim/cli/upgrade/utils.py#L75)

alberttorosyan (commented):
looks good! left a couple of comments (basically improve converted data accuracy)

gorarakelyan:
Some thoughts:
- Add tqdm to display a progress bar showing conversion progress
- Add audio, text, distributions plugins support
- What about recursively searching all the event files in deeper directories as well? Does it make sense? @mahnerak

alberttorosyan:
@gorarakelyan, regarding this:
> * What about recursively searching all the event files in deeper directories as well?

it is actually done, see this line of code: https://github.com/aimhubio/aim/blob/110b3baf72266a7cef2894ae58e8f5cc6084df2e/aim/cli/up/utils.py#L95

> * Add tqdm to display a progress bar showing conversion progress

Progress bar will definitely improve UX 👍 

> * Add audio, text, distributions plugins support

Initially we planned to support images and scalars, but I guess this goes back to times when we did not have support for other inputs.

aim/cli/up/utils.py
```diff
@@ -27,3 +32,119 @@ def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile):
...
+        return
+
+    try:
+        import tensorflow as tf
+        from tensorflow.python.summary.summary_iterator import summary_iterator
+    except ModuleNotFoundError:
+        logger.error('Cannot process TF events - could not import required modules.')
+        return
+
+    tf_logs_cache_path = f'{repo_inst.path}/tf_logs_cache'
```
mahnerak, line 61: (Very minor thing here) Please `path.join()` instead.

aim/cli/up/utils.py
```diff
@@ -27,3 +32,119 @@ def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile):
...
+                'run_hash': run.hash,
+                'events': {},
+            })
+        run_tf_events = run_cache['events']
+
+        events_to_process = []
+        events = [
+            os.path.join(root, file)
+            for root, dirs, files in os.walk(path)
+            for file in files if file.endswith('v2')
```
mahnerak, line 96: What about `tensorboard` event files not ending with `v2`?

aim/cli/up/utils.py
```diff
@@ -27,3 +32,119 @@ def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile):
...
+                        tf_tensors = value.tensor.string_val[2:]
+                        track_val = [
+                            Image(tf.image.decode_image(tf_tensor).numpy()) for tf_tensor in tf_tensors
+                        ]
+                        if len(track_val) == 1:
+                            track_val = track_val[0]
+                    elif plugin_name == 'scalars':
+                        track_val = value.simple_value
+                    else:
+                        # unsupported plugin type
```
mahnerak, line 140: Can we print a warning (only once) so user will know not all the data is being converted.

aim/cli/up/utils.py
devfox-se: how i missed that 🤦‍♂️

aim/cli/up/utils.py
devfox-se: Any idea how we can filter out only tf events? i thought v2 is the only good indicator

mahnerak: Example:
```
/mnt/logs/warp/few_rte-init-1614794261/log/train/:
total 72K
drwxr-xr-x 4 mahnerak mahnerak 4.0K Mar  3  2021 ..
drwxr-xr-x 2 mahnerak mahnerak 4.0K Mar  3  2021 .
-rw-r--r-- 1 mahnerak mahnerak  61K Mar  3  2021 events.out.tfevents.1614794275.4c0f0e71cd2f

/mnt/logs/warp/few_rte-init-1614794548/log/train/:
total 72K
drwxr-xr-x 4 mahnerak mahnerak 4.0K Mar  3  2021 ..
drwxr-xr-x 2 mahnerak mahnerak 4.0K Mar  3  2021 .
-rw-r--r-- 1 mahnerak mahnerak  61K Mar  3  2021 events.out.tfevents.1614794561.4c0f0e71cd2f

/mnt/logs/warp/few_rte-init-1614794842/log/train/:
total 72K
drwxr-xr-x 4 mahnerak mahnerak 4.0K Mar  3  2021 ..
drwxr-xr-x 2 mahnerak mahnerak 4.0K Mar  3  2021 .
-rw-r--r-- 1 mahnerak mahnerak  61K Mar  3  2021 events.out.tfevents.1614794860.4c0f0e71cd2f

/mnt/logs/warp/few_rte-init-1614795145/log/train/:
total 72K
drwxr-xr-x 4 mahnerak mahnerak 4.0K Mar  3  2021 ..
drwxr-xr-x 2 mahnerak mahnerak 4.0K Mar  3  2021 .
-rw-r--r-- 1 mahnerak mahnerak  61K Mar  3  2021 events.out.tfevents.1614795159.4c0f0e71cd2f
```

devfox-se: will stick to `events.out.tfevents` pattern

mahnerak:
In `tensorboard`, if there multiple files in a same directory, all of them are joined into one experiment/run.
Do we handle this?

aim/cli/up/utils.py
devfox-se: good idea, will add it

devfox-se:
@mahnerak we decided create a separate Run for each directory, so every event file inside that directory will be parsed and added to the corresponding Run

mahnerak:
Here's what I mean:
https://colab.research.google.com/drive/1dH6FErHEMOi4_bI-LX5-blm7PKdIoRNS?usp=sharing

mihran113 (commented):
don't want to ruin the conversation, but @devfox-se can you please update the `CHANGELOG` as well?

aim/cli/convert/utils.py
```diff
@@ -0,0 +1,254 @@
...
+        - image
+
+    To make conversion process smooth ensure that your directory has the following structure
+
+    your_provided_log_dir/
+        ├> run_1/
+        │    ├> <tf_events_file_1>
+        │    └> <tf_events_file_2>
+        ├> group_1/
+        │    ├> <tf_events_file_3> (THIS EVNET WILL BE IGNORED)
```
mahnerak, line 27: Typo here

aim/cli/convert/utils.py
```diff
@@ -0,0 +1,254 @@
...
+        │        ├> <tf_events_file_10>
+        │        └> <tf_events_file_11>
+        ├> <tf_events_file_12> (THIS EVNET WILL BE IGNORED)
+        └> <tf_events_file_13> (THIS EVNET WILL BE IGNORED)
+
+    From the hierarchy example above you can see that
+    "your_provided_log_dir/group1/tf_events_file_3",
+    "your_provided_log_dir/tf_events_file_12" and
+    "your_provided_log_dir/tf_events_file_13"
+    event files will be ignored since the converter threats them as unorganized event files.
```
mahnerak, line 47: In contrast to those mentioned, `tf_events_file_8` is _not_ ignored, right?
Can you please elaborate in comments what happens to that event file.
It takes default context, I guess?

aim/cli/convert/utils.py
```diff
@@ -0,0 +1,254 @@
...
+            return os.path.dirname(current_path)
+        elif current_path in ('', '.', '/'):
+            return current_path
+        return get_parent(os.path.dirname(current_path), level - 1)
+
+    tf_logs = os.path.abspath(tf_logs)
+    run_dir_candidates = set()
+    for root, dirs, files in os.walk(tf_logs):
+        for file in files:
+            if 'events.out.tfevents' not in file:
```
mahnerak, line 89: Maybe `os.path.basename(file).startswith(events.out.tfevents)`?

aim/cli/convert/utils.py
```diff
@@ -0,0 +1,254 @@
...
+    From the hierarchy example above you can see that
+    "your_provided_log_dir/group1/tf_events_file_3",
+    "your_provided_log_dir/tf_events_file_12" and
+    "your_provided_log_dir/tf_events_file_13"
+    event files will be ignored since the converter threats them as unorganized event files.
+
+    In case the converter finds unorganized event files in your hierarchy a warning message will be issued.
+
+    To make the converter process these events, consider re-structuring your directories so that it matches
+    the sample structure. (i.e. create a new directory and moving your unorganized events there)
```
mahnerak, line 52: + something like "Or set the `flat` flag to treat each directory as a distinct run."?

mahnerak (approved):
LGTM
Just couple of minor things to change.

aim/cli/convert/utils.py
devfox-se: eagle eye

aim/cli/convert/utils.py
devfox-se: I'll also add this

aim/cli/convert/utils.py
devfox-se: You're right, its empty context - I'll add it to docstring

alberttorosyan (approved):
looks great!
the `docstring` is really nice with the logs structure diagram ❤️

devfox-se:
> looks great! the `docstring` is really nice with the logs structure diagram ❤️

thank you:)

PR #1199, opened 2022-01-14
Feature/non strict writes

mahnerak:
Usually in higher level frameworks our users log the hyperparameter `dict` / `Namespace` / ... taken from third-party libraries. Such data may contain non-serializable-yet object thus yielding to errors when logging them.

For now it may be better option to write hyperparameters in non-strict mode: all the non-serializable objects will be stored like their string representations.


alberttorosyan (approved):
lgtm

Issue #1206, opened 2022-01-17
Numpy scalars logged with track() don't show up in list of metrics when running 'aim up' interface


mattaq31:
## 🐛 Bug

Numpy scalars logged with track() don't show up in list of metrics when running 'aim up' interface.

### To reproduce

Run the following Python code:

```
from aim import Run
import os
import random
import numpy as np
int_hash = random.randint(1,10000)
aim_session = Run(experiment='tst', run_hash=str(int_hash), system_tracking_interval=2)
for i in range(20):
    aim_session.track(i, name='test_loss', epoch=i)
    aim_session.track(np.array(i*2), name='test_numpy', epoch=i)

```
### Expected behavior

Both 'test_loss' and 'test_numpy' should show up in aim up interface.  Only 'test_loss' shows up in this case.  Numpy scalars are always ignored.
### Environment

- Aim 3.3.5
- Python 3.7
- Numpy 1.21.1
- Mac OS Monterey

### Additional context

Aim installed with Pip, rest of environment packages installed with Conda.

roubkar:
The fix has been shipped with `v3.4.0` release.

PR #1207, opened 2022-01-17
[fix] Track converted value type instead of the input value type.

alberttorosyan:
When saving tracked value type in run metadata, use the already
converted value instead of the original input value.
Added covering unit-test.


alberttorosyan:
resolves #1206

mihran113 (approved):
lgtm

Issue #1084, opened 2021-12-02
Store and retrieve all datetime objects and timestamps as UTC tz-aware times


alberttorosyan:
Currently run `created_at` is stored in `sqlite` db as UTC datetime
`finalized_at` value is stored in `rocksdb` as UTC datetime converted to timestamp. This is not correct since `datetime.datetime.timestamp()` assumes local tz if datetime is not timezone-aware. Need to fix this and store timestamp of tz-aware UTC datetime.
Once this is done, there will be an issue with existing repos: run duration calculation might be incorrect and diff will depend on tracking machine timezone. Need to provide rocksdb data migration which will fix the run finalized_at timestamps.

alberttorosyan:
Also meed to fix tracked value timestamp to be a UTC tz-aware datetime's timestamp as well

henrydavt:
Hello, I have been looking into this problem and I would suggests to change statement: https://github.com/aimhubio/aim/blob/main/aim/sdk/run.py#L70 to `time.time()` since `time()` always return timestamp in UTC

`time.time() - datetime.datetime.utcnow().timestamp() ~= 14400 seconds (4 hours)`

https://stackoverflow.com/questions/13890935/does-pythons-time-time-return-the-local-or-utc-timestamp

alberttorosyan:
Hi @henrydavt! Thanks for your comment!
Yes, using `time` module instead of `datetime` is an option. However the problem we have is a bit deeper, since we also have data stored as `datetime` objects (as UTC datetime).
Ideally we should store finalization time as `datetime` object (instead of timestamp) but at the moment we do not have `datetime` objects serialization for our storage.
The implementation of `datetime.timestamp()` is falling back to local if tz info is missing, that's the source of the issue. So the option we're considering now is to add tz info to the `datetime.datetime.utcnow()` before getting the timestamp.

PR #1216, opened 2022-01-19
Fix timezone issues

mihran113:
Change timestamps to be timezone aware instead of using local timezone


alberttorosyan (approved):
looks good

mihran113:
Resolves #1084

PR #1217, opened 2022-01-19
[feat] extra sys params tracking

devfox-se:


aim/sdk/utils.py
```diff
@@ -84,3 +84,56 @@ def check_types_compatibility(dtype: str, base_dtype: str, update_base_dtype_fn=
     if dtype == 'list' and any_list_regex.match(base_dtype):
         return True
     return False
+
+
+def get_installed_packages():
```
alberttorosyan, line 89: what about moving all these util functions into `aim/ext` module? looks to be completely independent from sdk stuff.

aim/sdk/run.py
```diff
@@ -269,6 +278,20 @@ def __init__(self, run_hash: Optional[str] = None, *,
...
+            if log_system_params:
+                system_params = {
+                    'packages': get_installed_packages(),
+                    'env_variables': get_environment_variables(),
+                    'git_info': get_git_info(),
+                    'executable': sys.executable,
+                    'arguments': sys.argv
+                }
+
+                if not system_params['git_info']:
```
alberttorosyan, line 290: I think leaving an empty dict here should be OK

aim/sdk/run.py
```diff
@@ -269,6 +278,20 @@ def __init__(self, run_hash: Optional[str] = None, *,
         ).subtree('seqs').subtree('chunks').subtree(self.hash)
 
         if not read_only:
+            if log_system_params:
```
mihran113, line 281: maybe we can put the `system_params` construction inside some utility function as well? as the `__init__` method of the `Run` class is pretty big and messy already by itself

mihran113 (approved):
everything else is fine, but left one comment there

aim/sdk/run.py
```diff
@@ -236,7 +244,8 @@ def __init__(self, run_hash: Optional[str] = None, *,
                  repo: Optional[Union[str, 'Repo']] = None,
                  read_only: bool = False,
                  experiment: Optional[str] = None,
-                 system_tracking_interval: Optional[int] = DEFAULT_SYSTEM_TRACKING_INT):
+                 system_tracking_interval: Optional[int] = DEFAULT_SYSTEM_TRACKING_INT,
+                 log_system_params: Optional[bool] = True):
```
alberttorosyan, line 248: lets make `False` by default at least for now. Without UI changes this will blow the Run/Metric explorer tables.

aim/sdk/utils.py
devfox-se: yeah moving to `aim/ext/resources/utils.py` is more convenient

aim/sdk/utils.py
```diff
@@ -84,3 +84,56 @@ def check_types_compatibility(dtype: str, base_dtype: str, update_base_dtype_fn=
...
+    import pkg_resources
+    packages = {i.key: i.version for i in pkg_resources.working_set}
+
+    return packages
+
+
+def get_environment_variables():
+    env_mask = ('sec', 'key', 'tok', 'pwd', 'pass')
```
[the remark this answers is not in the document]
devfox-se: will use full words

PR #1218, opened 2022-01-19
Improve error handling in the remote tracking 

### LINKED ISSUE #1201: remote tracking server error handling
Currently in case of exception on server side generic `RuntimeError` is raised in appropriate handler on client side.
This reduces ability to handle errors in higher-level SDK methods. For example `TreeView` might raise `KeyError` which is perfectly normal for `Run` object.

aramaim:
Change exception handling logic. Now client gets exact exception from server instead of generic `RuntimeError`.


mihran113 (approved):
@aramaim looks great! 
Just a note regarding the PR <-> issue linking: please edit the description and add a separate comment (`resolves #issueNO`), because the first option links PR with issue which we would like to avoid as when the PR is closed it'll automatically close the issue as well

aramaim:
resolves #1201

aim/ext/transport/server.py
```diff
@@ -42,46 +45,68 @@ class RemoteTrackingServicer(remote_tracking_pb2_grpc.RemoteTrackingServiceServi
...
+        try:
+            resource_handler = request.handler
+            client_uri = request.client_uri
+            if not self._verify_resource_handler(resource_handler, client_uri):
+                raise UnauthorizedRequestError()
+
+            del self.resource_pool[resource_handler]
+            return rpc_messages.ReleaseResourceResponse(status=rpc_messages.ReleaseResourceResponse.Status.OK)
+        except Exception as e:
+            # Should we clean ed resource_pool in case of exceptions? what if UnauthorizedRequest?
```
alberttorosyan, line 83: what if instead of `UnauthorizedRequest` we do not raise exception than catch it and handle but just return `ReleaseResourceResponse` with ERROR status and appropriate exception? so basically in LINE 78:
```
return rpc_messages.ReleaseResourceResponse(
    status=rpc_messages.ReleaseResourceResponse.Status.ERROR,
    exception=build_exception(UnauthorizedRequestError()),
)
```

alberttorosyan (commented):
looks great!
please remove all `RuntimeError` handling added for remote tracking. it should be easy to find by comment `"revisit once proper error handling is added"`

aim/ext/transport/server.py
aramaim: Thanks for notes. I guess `raise` more readable than `return` in this particular case. Plus with `Single Return` we can easily add some clean-up codes or easily change Exception safety of the methods.

alberttorosyan (approved):
lgtm

alberttorosyan (approved):
lgtm

PR #1221, opened 2022-01-20
(feat) matplotlib integration

devfox-se:
I don't think creating a new CustomObject for `matplotlib` is reasonable since we can convert it both to Plotly figure and to Image, thus I have made appropriate changes in both custom objects to support conversion


aim/sdk/objects/image.py
```diff
@@ -56,6 +56,8 @@ def __init__(self, image, caption: str = '', format='png', quality=85, optimize=
             self._from_tf_tensor(image, params)
         elif inst_has_typename(image, ['numpy', 'array']):
             self._from_numpy_array(image, params)
+        elif inst_has_typename(image, ['Figure.matplotlib.figure']):
```
mihran113, line 59: @devfox-se I guess this need to be changed to `['Figure', 'matplotlib', 'figure']` based on `inst_has_typename` function implementation

devfox-se: Will change it only for `Image` class

aim/sdk/objects/figure.py
```diff
@@ -5,7 +5,9 @@
 
 @CustomObject.alias('aim.figure')
 class Figure(CustomObject):
-    """Figure object used to store plotly figures in Aim repository.
+    """
+    Figure object can be used for storing Plotly or Matplotlib figures into Aim repository.
+    Core functionality is based on Plotly.
```
alberttorosyan, line 10: Please add `matplotlib.figure` to `Args` section as well.

aim/sdk/objects/figure.py
```diff
@@ -16,7 +18,9 @@ class Figure(CustomObject):
     def __init__(self, obj):
         super().__init__()
 
-        if inst_has_typename(obj, ['Figure', 'BaseFigure']):
+        if inst_has_typename(obj, ['Figure.matplotlib.figure']):
```
alberttorosyan, line 21: @mihran113's comment applies here as well; should be `inst_has_typename(image, ['Figure', 'matplotlib', 'figure'])`

aim/sdk/objects/image.py
```diff
@@ -222,6 +224,18 @@ def _from_tf_tensor(self, tensor, params):
...
+        try:
+            assert hasattr(figure, 'savefig')
+        except AssertionError:
+            raise ValueError('Cannot convert from matplotlib figure.')
+
+        buffer = BytesIO()
+        figure.savefig(buffer)
+        buffer.seek(0)
+
+        return self._from_pil_image(PILImage.open(buffer), params)
```
alberttorosyan, line 237: It's OK for now to use `PILImage` as intermediate object to convert to png/jpeg. However it would be good to write `buffer` directly and make sure the appropriate metadata (`'source'`, `'format'`, etc.) is set correctly. Please open a Github issue for that.

alberttorosyan (commented):
looks good! left a couple of comments.

alberttorosyan (approved):
lgtm

PR #1232, opened 2022-01-21
Enhance response data process(error) handling

arsengit:
- Wrapped all ui components in pages directory into `ErrorBoundary` component
- Added `exception handler` for response data process in explorers models 
- Added `Redirect` route in case of bad url


arsengit:
Resolves #1236

rubenaprikyan (approved):

PR #1242, opened 2022-01-22
Add experiment name default value in images explore table

VkoHov:
- Added default value as "default". If there isn't an experiment name tracked.


rubenaprikyan (approved):
lgtm

Issue #1198, opened 2022-01-14
Ability to hide system metrics from table


roubkar:
## 🚀 Feature

Add ability to toggle all system metrics columns visibility from table

### Motivation

When you don't want to see system metrics columns in runs/params/scatters tables there is no easy way to hide them, it would be great to have an option for hiding the system metrics column.

### Pitch

Add a toggle action to the `Manage Columns` popover which will hide/show system metrics columns.

### Additional context

Discussion: https://github.com/aimhubio/aim/discussions/1023

gloryVine:
@arsengit This looks great, it will remove a lot of clutter!

On a related note, currently my issue is that I have a lot of parameters which each appear as their own column so I can compare better across runs.
- In practice, I only ever need to see certain subgroups of these parameters, and I think it would be of great help if one were able to have something like 'meta columns' which when collapsed would collapse the multiple columns belonging to that meta column.
- For some reason, it is possible to increase column width arbitrarily, whereas it is not possible to minimize column size beyond the point where this would mean cutting off the column name. Again, in practice, it would be extremely helpful to be able to arbitrarily resize column widths. For example, I tend to have very long run-names, but at any given moment I might not be interested in seeing those completely.

It would be great if you could let me know whether any of these two features is already available / planned, or whether I should start a new feature request.

Thanks a lot again for the possibility to hide system metrics.

arsengit:
Hey @gloryVine, thanks a lot for sharing your experience!

So one feature is the ability to manage columns by their group (e.g. hide all params columns, hide all metrics columns)
Another feature request can come out from here: Add ability to group table columns.

As for minimizing column width: this is a bug.

It would be super helpful if you could open feature request/bug report for these items with additional context 🙏

We also have `Show table diff` action button in `Manage Columns` popover which will hide all the columns where each row has the same value.

<img width="756" alt="Screen Shot 2022-01-28 at 16 10 28" src="https://user-images.githubusercontent.com/44676488/151544755-22090ff0-5e7b-4d9d-b5d9-88f833985cd1.png">

Would love to invite you to join [Aim community slack](https://slack.aimstack.io) where you can ask questions, suggest ideas, share your experience and provide feedback 🙂

PR #1250, opened 2022-01-25
Add ability to hide system metrics from table

arsengit:
- Added `Hide System Metrics` button in `Manage Columns` popover 
- Added ability to hide system metrics from explorer tables
- Added `xSmall` size for `Button` component


arsengit:
Resolves #1198

VkoHov (approved):
lgtm!

aim/web/ui/src/services/models/explorer/createAppModel.ts
```diff
@@ -340,6 +340,7 @@ function createAppModel(appConfig: IAppInitialConfig) {
           config.table = {
             resizeMode: ResizeModeEnum.Resizable,
             rowHeight: RowHeightSize.md,
+            hideSystemMetrics: true,
```
KaroMourad, line 343: let's add this config state also for Metrics table

aim/web/ui/src/services/models/explorer/createAppModel.ts
```diff
@@ -2294,6 +2282,28 @@ function createAppModel(appConfig: IAppInitialConfig) {
...
+          sameValueColumns: string[];
+        },
+        tableColumns: ITableColumn[],
+        configData: IAppModelConfig | any,
+      ): void {
+        if (updateTableTimeoutId) {
+          clearTimeout(updateTableTimeoutId);
+        }
+
+        updateTableTimeoutId = setTimeout(() => {
```
KaroMourad, line 2297: here let's use `window.setTimeout` and fix the type of `updateTableTimeoutId: ReturnType<typeof setTimeout>;`
to number

aim/web/ui/src/services/models/explorer/createAppModel.ts
arsengit: @KaroMourad Metrics table haven't system metrics

aim/web/ui/src/services/models/explorer/createAppModel.ts
arsengit: @KaroMourad  There are no difference, except that windows.setTimeout works only in browser environment

aim/web/ui/src/services/models/explorer/createAppModel.ts
arsengit: In Metrics table the Hide/Show button isn't rendered

KaroMourad: @arsengit the thing is we can have the same table config for all of them to easily integrate for every app model, we need it for app model unification @roubkar  what do you think ?

aim/web/ui/src/services/models/explorer/createAppModel.ts
KaroMourad: @arsengit we can use it only on browser env that's why we can use `window.setTimeout` here

aim/web/ui/src/services/models/explorer/createAppModel.ts
roubkar: we can think of this during base explorer implementation, for now this is ok I think

PR #1252, opened 2022-01-25
[feat] Aim repo automatic reindexing

alberttorosyan:
Added `RepoIndexManager` class to handle repo re-indexing (online & offline).
The `start_indexing_thread` method will constantly check for stalled runs
in a separate worker thread. Once stalled runs found, they will be indexed
starting from least recent run. There's a 2 sec. interval between 2 consecutive
indexing iterations to ensure index db lock is released in between.
Once all runs are indexed, sleep interval is increased up to 10 sec.

Refactored `aim reindex` command to re-use `RepoIndexManager` class.
Enhanced command to do second pass over runs in progress.


aim/sdk/index_manager.py
```diff
@@ -0,0 +1,180 @@
...
+    from aim.sdk.repo import Repo
+
+from aim.sdk.run import Run
+from aim.storage.rockscontainer import RocksContainer
+
+logger = logging.getLogger(__name__)
+
+
+class RepoIndexManager:
+    index_manager_pool = {}
```
mihran113, line 26: why do we need a pool though? isn't the manager going to be alive through the whole `aim up` up time, and will we have more than one manager inside the same process?

aim/sdk/index_manager.py
```diff
@@ -0,0 +1,180 @@
...
+        return run_hashes
+
+    def _in_progress_runs(self) -> List[str]:
+        progress_dir = os.path.join(self.repo.path, 'meta', 'progress')
+        runs_with_progress = os.listdir(progress_dir)
+        run_hashes = sorted(filter(lambda r: self._is_run_in_progress(r), runs_with_progress),
+                            key=lambda r: os.path.getmtime(os.path.join(progress_dir, r)))
+        return run_hashes
+
+    def _run(self, run_hash):
```
mihran113, line 174: what's the point wrapping the `Run` inside a private method?

aim/sdk/index_manager.py
alberttorosyan: @mihran113 the point is to have a single instance of `RepoIndexManager` per repo. Regardless how many times it's needed (for example API call to check indexing status should not create a new instance).
But your point is valid, since at the moment we do not have a use-case with more than one `Repo`.
Maybe worth considering a `Singleton`? 🤔

mihran113: The other option is to add `IndexManager` instance to the members of `Project` class, and reuse it whenever needed

alberttorosyan: @mihran113 that would create a dependency between cli `reindex` command and `web` module.

aim/sdk/index_manager.py
alberttorosyan: just to make the line shorter in a couple of places 😄
`self._run(run_hash)`
vs.
`Run(run_hash, repo=self.repo, system_tracking_interval=None)`

aim/sdk/index_manager.py
mihran113: let's keep it now as it is then

aim/sdk/index_manager.py
mihran113: okay, now seeing it side by side makes sense 😆

mihran113 (approved):
lgtm

alberttorosyan:
Resolves #1284

Issue #1228, opened 2022-01-21
Single rpc call in Run.track()


alberttorosyan:
Current implementation of `Run.track()` makes multiple write operation on `TreeView` objects. This can potentially lead to the data inconsistency, when exception is raised in the middle of `track()`. With the `aim` remote tracking the risk of such cases is high (for example, network connection issues).
Need to enhance the `track()` implementation to make a single RPC call with the batch of `TreeView` method calls.

PR #1254, opened 2022-01-25
Atomic track for remote tracker

aramaim:
I added dedicated queue for processing non blocking writes and made read blocking. As well as I added logic to combine all write instruction from `run.track()` and send combined writes as a single rpc call.


aramaim:
resolves #1228

aim/sdk/repo.py
```diff
@@ -739,3 +741,11 @@ def close(self):
     @property
     def is_remote_repo(self):
         return self._client is not None
+
+    @contextmanager
+    def atomic_track(self):
```
alberttorosyan, line 746: maybe move to `Run`?

aim/sdk/run.py
```diff
@@ -442,14 +444,15 @@ def _track_impl(
...
+        with self.repo.atomic_track():
+            self.meta_run_tree['traces', ctx.idx, name, 'last'] = val
+            self.meta_run_tree['traces', ctx.idx, name, 'last_step'] = step
 
-        # TODO perform assignments in an atomic way
-        seq_info.val_view[step] = val
-        seq_info.epoch_view[step] = epoch
-        seq_info.time_view[step] = track_time
-        seq_info.count = seq_info.count + 1
+            # TODO perform assignments in an atomic way  # Ask AT, actually whole method should be atomic?
```
alberttorosyan, line 451: yes 😃

alberttorosyan (changes requested):

aim/sdk/run.py
aramaim: fixed

aim/sdk/repo.py
aramaim: Not sure if it is a good idea, because in that case we have to make `client` "public" in `repo` class. Or provide almost the same interface in repo as well.

aim/ext/transport/client.py
```diff
@@ -62,13 +84,43 @@ def message_stream_generator():
             for chunk in stream:
                 yield rpc_messages.InstructionRequest(message=chunk)
 
+        # TODO [AD] make this atomic in case of parallel write from different thread?
+        # TODO [AD] make queue per run? do we need it?
+        self._queue.wait_for_finish()
```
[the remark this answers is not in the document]
aramaim: I'll implement that logic in the scope of #1229 issue.

aim/ext/transport/client.py
```diff
@@ -62,13 +84,43 @@ def message_stream_generator():
...
+                    version='0.1',
+                    client_uri=self.uri,
+                    message=chunk
+                )
+
+        response = self.remote.run_write_instructions(message_stream_generator())
+        if response.status == rpc_messages.WriteInstructionsResponse.Status.ERROR:
+            raise_exception(response.header.exception)
+
+    def init_tracking(self):
```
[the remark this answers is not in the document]
aramaim: start_instructions_batch/flush_instructions_batch

mihran113:
Haven't gotten through all the code yet, but here are some issues found while testing on this branch:
we've lost the handling of the `KeyboardInterrupt` when there are still some tasks pending in the `metric_tracking` queue after double `ctrl-C`s after that the process can be killed with `ctrl-C`. Also tasks were stuck infinitely in the `metric_tracking` queue after double `ctrl-C`. 
To reproduce:
I've used https://github.com/aimhubio/aim/blob/main/examples/pytorch_lightning_track.py this example for the testing. Just added `repo='aim:0.0.0.0:53800'` argument for the `AimLogger` and run the server somewhere else.

aramaim:
thanks @mihran113 for pointing this out! I've pushed fix.

aim/sdk/run.py
```diff
@@ -100,6 +108,7 @@ def _close(self) -> None:
             return
         self.finalize_system_tracker()
         self.finalize_run()
+        self.finalize_remote_tracking_queue()
```
mihran113, line 111: this will do the trick for now, but I guess it's not the best idea for the run to wait until the tracking queue is fully empty, as there might be other runs in progress as well (in the same queue). Maybe put a `Todo` item for furhter improvement?

mihran113 (approved):
left a minor note there :) other than that, looks pretty good

aim/sdk/run.py
aramaim: make sense. done

alberttorosyan (approved):
looks good

PR #1264, opened 2022-01-28
[fix] handling omegaconf in nested structure

### LINKED ISSUE #1239: Omegaconf not supported by the pytorch_lightning logger
## 🐛 Bug
The config file of `hparam`, in the format of the `OmegaConf` object, cannot be saved by aim's `pytorch_lightning` logger properly. It will be stored as a string instead of a `dict`.

It would work only if I manually convert the config to a `dict`.

### Environment
- Aim: 3.3.5
- Python: 3.9.7
- pip: 21.3.1
- OS:  Linux
- pytorch-lightning: 1.5.8
- omegaconf: 2.1.1

devfox-se:
Fixes #1239

### DIFF
--- a/aim/storage/treeutils_.pyx
+++ b/aim/storage/treeutils_.pyx
@@ -30,8 +30,7 @@ def encode_non_native_object(
 ):
     if omegaconf_installed and isinstance(obj, Container):
         obj = OmegaConf.to_container(obj, resolve=True)
-        for key, val in obj.items():
-            yield from unfold_tree(val, path=path + (key,), unfold_array=unfold_array, depth=depth, strict=strict)
+        yield from unfold_tree(obj, path=path, unfold_array=unfold_array, depth=depth, strict=strict)
     else:
         raise TypeError(f'Unhandled non-native value `{obj}` of type `{type(obj)}`.')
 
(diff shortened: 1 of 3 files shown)


aim/storage/treeutils_.pyx
```diff
@@ -12,6 +12,28 @@ from aim.storage.object import CustomObjectBase, CustomObject
...
 
+try:
+    from omegaconf import OmegaConf, Container
+except ModuleNotFoundError:
+    omegaconf_installed = False
+else:
+    omegaconf_installed = True
+
+
+def encode_non_native_object(
```
alberttorosyan, line 23: lets keep the `treeutils` generic.
instead we can add a `CustomObject` which will wrap OmegaConf Container and use its' dict-like interface to serialize the data.
@mahnerak thoughts/suggestions?

alberttorosyan (changes requested):

aim/storage/treeutils_.pyx
devfox-se: As discussed earlier - non native converter is moved to another file

alberttorosyan (approved):
lgtm

PR #1266, opened 2022-01-31
 [fix]  Add ability to change value prop of toggle button component dynamically

Hamik25:
- Add value prop change listener to handle the value in toggle button component dynamically

### DIFF
--- a/aim/web/ui/src/components/kit/ToggleButton/ToggleButton.tsx
+++ b/aim/web/ui/src/components/kit/ToggleButton/ToggleButton.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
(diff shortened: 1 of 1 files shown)


arsengit (approved):

roubkar:
2 questions here:
- does this component need a state at all? (seems `value` prop can do it all)
- if there's a need for a state maybe we can also add prop and state equality check before updating?
```javascript
useEffect(() => {
    if (value !== state) {
        setState(value);
    }
}, [value]);

File: aim/web/ui/src/components/kit/ToggleButton/ToggleButton.tsx [hunk 1]
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
```
roubkar, line 1: seems no need for `useEffect`

Issue #1194, opened 2022-01-13
UI background is partially transparent in colab


rubenaprikyan:
## 🐛 Bug

UI background is partially transparent in colab
Params explorer's background is transparent when using dark mode of browser

### Expected behavior
Params explorer's background should not be transparent

### Environment

- Aim Version 3.3.3
- in colab

### How to solve
Colab is setting it's colors to the body and children components  are inherited , need to set backgrounds for pages or set background colors for body

rubenaprikyan:
Resolves #1194

PR #1274, opened 2022-02-02
[fix] Change background transparency in colab when using dark mode of system

rubenaprikyan:
Change background transparency in `colab` when using dark mode of system


rubenaprikyan:
Resolves #1194

roubkar (approved):

Issue #1269, opened 2022-02-02
Aim server give errors after deleting all runs


mikel-brostrom:
## 🐛 Bug

1. I set up a remote aim server
2. Train a few models and log data to aim
3. I deleted all the runs thrugh the UI
4. Aim start complaining:
`aimrocks.errors.RocksIOError: b'IO error: No such file or directory: While opening a file for sequentially reading: /xxx/aim_server/.aim/meta/chunks/a93447ab850242f588fd311d/CURRENT: No such file or directory'`

### Environment

- aim                     3.4.1
- Python 3.7.9
- pip                     21.2.4
- Linux 18.04

mihran113:
Hey @mikel-brostrom, looking at this right now, will provide updates shortly after reproducing the issue on our side.
Thanks for opening the issue :)

mihran113:
@mikel-brostrom thanks again for pin-pointing the issue, I've reproduced this on my side and started already working on it. In the meantime, can I ask if this is a blocker for you?
By the way in this case(deleting all the runs), you can just re-initialize the repository calling `aim init` on the same repo.

mikel-brostrom:
Not really a blocker. I was mostly testing the functionality and came across this bug. Glad I could help!

mihran113:
> Not really a blocker. I was mostly testing the functionality and came across this bug. Glad I could help!

Thanks for a quick response. If it's not a blocker, then I'll change the milestone to v3.5.0 as it's planned to be released this week.

mihran113:
Hi @mikel-brostrom. This fix has been shipped with v3.5.0. Please have a try and let us know if everything works fine.

PR #1275, opened 2022-02-02
Fix stalled runs deletion

mihran113:
Also added check to identify if run is still in progress and changed `repo` methods(delete, copy, move) to not stop execution in case of failure and try to handle remaining runs as well.


mihran113:
Resloves #1269

alberttorosyan (approved):
lgtm

PR #1283, opened 2022-02-04
Fix of Run class creates weird folder issue

aramaim:
Avoid .aim folder creation for remote tracker when Repo object is created implicitly


aramaim:
resolves #1283

aim/sdk/repo.py
```diff
@@ -216,6 +216,8 @@ def rm(cls, path: str):
 
     @classmethod
     def check_repo_status(cls, path: str) -> RepoStatus:
+        if cls.is_remote_path(path):
```
devfox-se, line 219: maybe this could be a @property

devfox-se (approved):

Issue #1285, opened 2022-02-04
Fix system metric columns pinning functionality and grouping column order


arsengit:
## 🐛 Bug

System metric columns pinning functionality and grouping column order doesn't work properly.

### To reproduce

To reproduce system metrics bug, go to metric explore page (runs/params/scatters) and pin (left/right) system metrics column.
For grouping column order, pin to left any table column and apply grouping.

PR #1286, opened 2022-02-04
Fix system metric columns pinning functionality from table

arsengit:
- Fixed system metric columns pinning functionality from table
- Fixed `Grouping` column order


arsengit:
Resolves #1285

roubkar (approved):

VkoHov (approved):
lgtm!

Issue #1294, opened 2022-02-06
Activity heatmap doesn't paint the last day if it's Sunday


rubenaprikyan:
## 🐛 Bug

Activity heatmap doesn't paint the last day if it's Sunday

### To reproduce

Open home page on Sunday, it will show previews day as the latest cell in heatmap.

### Expected behavior

Activity heatmap should paint the current day on Sunday, if it's the latest cell in a heatmap 

### Environment

- Aim Version 3.5.0

PR #1295, opened 2022-02-06
[fix] Fix issue with displaying current day activity cell on week's first day

rubenaprikyan:
Fix issue with displaying current day activity cell on week's first day


rubenaprikyan:
Resolves #1294

arsengit (approved):

PR #1298, opened 2022-02-07
[feat] Unification of Run Custom Object APIs

VkoHov:
- Added `RangePanel` initial value calculation.
- Change property names in UI. Based on Run custom object API change.
- Add dynamic validation pattern usage both in images explore and run single page `RangePanel`

BE changes:
- Added common utilities for CustomObject-based APIs.
- Added class to auto-generate APIs based on the API config class.
- Adjusted existing object implementations (Image, Audio, etc.) and sequence classes.

API modifications (breaking change)
- Removed `index|record_range` and `index|record_slice`
- Removed `calc_ranges` input
- Total ranges are always calculated
- Both total and used ranges added to response schema.


alberttorosyan:
resolves #1300

rubenaprikyan (approved):
lgtm

mihran113 (approved):
looks absolutely fantastic 🎉

alberttorosyan:
> @alberttorosyan regarding the discussion we had today: maybe it's worth moving the endpoint registration part to `runs/views.py`, so all the endpoints can be accessible and visible from the same file

@mihran113 yes, will do that later today.

arsengit (approved):

PR #1302, opened 2022-02-08
Fix issues for queries on remote server

mihran113:
- Avoid structured data caching when querying on remote
 - Change args encoding/decoding for resource request to be able to encode `NoneType` objects as well
 - Fix `keys()` and `items()` methods for `ProxyTreeView`.


aim/ext/transport/message_utils.py
```diff
@@ -10,6 +10,22 @@
 Message = Union[rpc_messages.ResourceRequest, rpc_messages.ResourceResponse]
 
 
+def pack_args(tree: Iterator[Tuple[bytes, bytes]]) -> bytes:
```
alberttorosyan, line 13: tree -> args?

aim/storage/treeview.py
```diff
@@ -89,6 +89,13 @@ def __setitem__(
     ):
         self.set(path, value, strict=True)
 
+    @abstractmethod
+    def keys_eager(
```
alberttorosyan, line 93: please add a TODO to revisit this approach.
do not want to do it now but after @mahnerak finishes the cython bindings implementation to avoid conflicts as much as possible.

alberttorosyan (approved):
left a couple of comments.
the rest looks good

aim/storage/treeview.py
mihran113: sure

aim/ext/transport/message_utils.py
mihran113: done ✅

PR #1305, opened 2022-02-09
[fix] live updated data loss after triggering endless scroll

VkoHov:
Fixed live updated data loss after triggering endless scroll issue


KaroMourad (approved):
lgtm

roubkar (approved):

Hamik25 (approved):

PR #1325, opened 2022-02-13
[doc] guides for custom objects

devfox-se:


alberttorosyan (commented):
content looks good. Just thinking about the placement of this files. maybe better to move to 
`v2/quick_start`?
@gorarakelyan what do you think?

PR #1335, opened 2022-02-15
Cython bindings

mahnerak:
- Initiate BASELINE rebuild by clearing the baseline file to continuously tracking the performance without forcing them match the previous baselines.
  - _BASELINE is going to recalculated after PR approval, right before merging the changes_.
  - Printing Performance Baselines. _I prefer to keep this change_ and implement more organized reports for performance tests.
- Using cython-optimized `aimrocks` version.
- Installing Cython before installing aim for tests/perf tests.
- Cython Native Bindings:
  - Optimized encoding,
  - Hashing,
  - Reorganizing Cython scripts,
  - Utils refactoring,
  - Bindings for Array View + Tree View,

I think we better off merge-squashing this time.


alberttorosyan (approved):
lgtm

PR #1348, opened 2022-02-15
Cython Bindings / Utils

mahnerak:
* Moving everything else other than type annotations from `types.py` to `utils.pyx`,
* Removing some (generic) type annotations to make it easier to optimize with Cython,
* Wrapper classes for `KeysIterator` and `ValuesIterator`.

### DIFF
--- a/aim/storage/utils.py
+++ b/aim/storage/utils.py
@@ -1,3 +1,5 @@
+from copy import deepcopy
+
 class ArrayFlagType:
     def __repr__(self):
         return "<ArrayFlag>"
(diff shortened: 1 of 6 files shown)


aim/storage/utils.py
```diff
@@ -1,3 +1,30 @@
+from copy import deepcopy
+
+
+class KeysIterator:
```
alberttorosyan, line 4: what's the purpose of the Iterator classes? is it something to be used in next PRs?

mahnerak: Yes! There will be fast Cython implementations for `ItemsIterator`s always yielding tuples of `(key, value)`s, and there is no point of optimizing them as well. `RocksDB` always retrieves values under the hood. Maybe we'll optimize `KeysIterator` after `RocksDB` does.

alberttorosyan (approved):

PR #1349, opened 2022-02-16
Cython Bindings / Array View + Tree View

mahnerak:
* Typed arrays,
* Removed some annotations:
  * Generics are not supported by Cython,
  * `@abstractmethod`s stop us from using `cdef`-ed classes as interfaces,
* Moved implementations to a separate file: Now `arrayview` is only for interface.
* Cython declarations for `TreeArrayView`, `ArrayView`, `TreeView`. Compiling them with Cython.


aim/storage/treearrayview.py
```diff
@@ -0,0 +1,185 @@
...
+
+from itertools import islice
+
+from typing import Any, Iterator, List, Tuple, Union
+from aim.storage.treeview import TreeView
+
+from aim.storage.arrayview import ArrayView
+
+
+class TreeArrayView(ArrayView):
```
alberttorosyan, line 11: need to change signature for `ProxyTree` as well (for remote tracking).

aim/sdk/run.py
```diff
@@ -645,9 +645,9 @@ def _read_sequence_info(self, ctx, name):
         seq_info = self.sequence_info[sequence_selector]
 
         assert not seq_info.initialized
-        seq_info.val_view = self.series_run_tree.subtree(sequence_selector).array('val').allocate()
-        seq_info.epoch_view = self.series_run_tree.subtree(sequence_selector).array('epoch').allocate()
-        seq_info.time_view = self.series_run_tree.subtree(sequence_selector).array('time').allocate()
+        seq_info.val_view = self.series_run_tree.subtree(sequence_selector).array('val', dtype='double').allocate()
```
alberttorosyan, line 648: this function is meant for generic sequences. we cannot be sure that `dtype='double'` is the correct choice.
though looking into the implementation of `TreeArrayView` it is only used to convert to numpy...

mahnerak: This is my fault. Yes, we have to figure out how to treat series / sequences as typed arrays.
Will roll back this one.

alberttorosyan (approved):
lgtm

PR #1372, opened 2022-02-21
Fix metaclass conflicts for pydantic models (py 3.6)

mihran113:
Removed generic support for custom object views


alberttorosyan (approved):
lgtm

PR #1381, opened 2022-02-22
[fix] Fix rendering issue on Select form search suggestions list

arsengit:
- Fixed select form search suggestions list

### DIFF
--- a/aim/web/ui/src/utils/app/getParamsSuggestions.ts
+++ b/aim/web/ui/src/utils/app/getParamsSuggestions.ts
@@ -1,0 +1,24 @@
+import { IProjectsModelState } from 'types/services/models/projects/projectsModel';
+
+/**
+ * @returns memoized params suggestions from projectsData
+ */
+
+export function getParamsSuggestions(data: IProjectsModelState) {
+  let list: string[] = [];
+  if (data?.params) {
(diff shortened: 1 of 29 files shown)


arsengit:
Resolves #1384

aim/web/ui/src/utils/app/getParamsSuggestions.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @returns memoized params suggestions from projectsData
+ */
+
+export function getParamsSuggestions(data: IProjectsModelState) {
+  let list: string[] = [];
+  if (data?.params) {
+    Object.keys(data?.params).forEach((option: any) => {
+      if (option) {
+        list.push(`run.${option}`);
+        if (data.params) {
```
Hamik25, line 13: I think here can be used an operator and one if statement with two conditions. data?.params[option] can be changed to data.params[option] if data hasn't param property or param prop is falsely value this scenario will not be executed.

Hamik25 (approved):

rubenaprikyan (approved):
lgtm!

rubenaprikyan (changes requested):
Update CHANGE LOG to  **Display `None` values  in group config column.**

Issue #1365, opened 2022-02-18
Improve documentation after restructuring


alberttorosyan:
## 📚 Documentation

The new structure of docs has clean structure and organized in a way to make it easy to get useful information regardless how deep is your knowledge of Aim.
However, certain areas can be improved further. Here is the list of what can be improved (feel free to add more items)

### TODO
- [ ] `understanding/overview.md` Put a components diagram at the bottom of the page.
- [ ]  `understanding/data_storage.md` Add data flow diagram (from tracking to UI).
- [ ]  Add `Glossary` section.
- [ ] Tune cross-links
- [ ] Add `Archive runs` in manage runs page
- [ ] Document querying of all media and object
- [ ] Document system resource usage and runs other attributes in `configuring runs` section
- [x] Document Aim UI concepts and guides

PR #1382, opened 2022-02-22
[doc] Document Aim UI concepts and guides

roubkar:
Added initial version of Aim UI docs to cover the usage of pages


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1365

Issue #1386, opened 2022-02-22
Fix font styles in Aim UI


arsengit:
##  🛠

Make font styles lighter across the project to provide users better reading experience.

PR #1387, opened 2022-02-22
[fix] Fix font styles in Aim UI

arsengit:
- Shifted down each of font families weights by 100
- Fixed font styles in `Home` page


arsengit:
Resolves #1386

KaroMourad (approved):

Issue #1389, opened 2022-02-22
Param name doesn't formatting in select forms


VkoHov:
## 🐛 Bug

In Params and Scatters explores param names keep `__example_type__` substrings.
`
### To reproduce

1.  Go to Params or Scatters explores
2. Open `SelectForm` dropdown

<img width="1256" alt="Screen Shot 2022-02-23 at 00 15 44" src="https://user-images.githubusercontent.com/43929589/155219873-fcdbe620-b595-4afd-87d3-1b9cce37340a.png">

### Expected behavior

Have to remove  `__example_type__` substrings from param name

### Environment

- Aim Version 3.6.0

PR #1390, opened 2022-02-22
[fix] Remove example type substring from param name

VkoHov:
- Remove `__example_type__` substring from param names both in Params and Scatter explores


VkoHov:
Resolves [#1389](https://github.com/aimhubio/aim/issues/1389)

rubenaprikyan (approved):
lgtm!

arsengit (approved):

PR #1393, opened 2022-02-23
[doc] Add documentation for Runs Management on UI.

rubenaprikyan:
- Added documentation for Runs Management on UI.


roubkar:
Maybe also worth mentioning navigation from table to Single Run Page?

PR #1400, opened 2022-02-24
[doc] Add tags page documentation.

Hamik25:


Hamik25:
Added tags page docs with topics

- Create tag
- Attach tag
- Update attached tags
- Edit tag
- Delete tag
- Used in overlay

PR #1406, opened 2022-02-25
[doc] run config docs

devfox-se:


alberttorosyan (commented):
lgtm.
@devfox-se just need to add `configure_runs.md` to the index and make sure the page appears in the built HTML version of docs.

alberttorosyan (approved):

Issue #1415, opened 2022-02-27
aim convert mlflow --experiment fails for experiment id, works for experiment name


luisoala:
## 🐛 Bug

doing

`$ aim convert mlflow --tracking_uri 'file:///Users/aim_user/mlruns' --experiment 61`

as described here https://aimstack.readthedocs.io/en/latest/quick_start/convert_data.html#show-mlflow-logs-in-aim

fails with the following error

![Screenshot from 2022-02-27 02-33-17](https://user-images.githubusercontent.com/26168435/155864827-dc7f3acb-0c79-4fab-9c79-a599f1a954ab.png)

using the experiment name instead of the experiment id

![Screenshot from 2022-02-27 02-33-55](https://user-images.githubusercontent.com/26168435/155864887-63c19423-865e-4540-bfb7-c034e123af80.png)

i.e.

`$ aim convert mlflow --tracking_uri 'file:///Users/aim_user/mlruns' --experiment 'ai-vengers-collab'` 

works:

![Screenshot from 2022-02-27 02-31-46](https://user-images.githubusercontent.com/26168435/155864881-03434a11-68f8-47e3-90e3-13465cbe86b4.png)

### To reproduce

see above

### Expected behavior

convert the experiment by ID

### Environment

- Aim Version 3.6
- Python 3.8.1
- pip3
- Ubuntu 20.04.3 LTS

gorarakelyan:
Hey @luisoala, thanks for reporting the issue!
@devfox-se could you please take a look at this?

devfox-se:
Thanks for reporting this @luisoala, will take a look soon!

alberttorosyan:
Hey @luisoala! We've released `v3.6.2` containing the fix for mlflow converter. Please check it out and let me know if there are any issues.

devfox-se:
@luisoala Hi, have you had a chance to test this?:)

devfox-se:
Closing due to inactivity, feel free to reopen in case this still persists.

PR #1420, opened 2022-02-28
[fix] mlflow conversion with experiment id/name

devfox-se:


devfox-se:
This resolves #1415

aim/cli/convert/processors/mlflow.py
```diff
@@ -41,11 +41,17 @@ def parse_mlflow_logs(repo_inst, tracking_uri, experiment):
...
 
+    experiment = kwargs.get('experiment_id') or kwargs.get('experiment_name')
+    get_experiment_by_name = kwargs.get('experiment_name') == experiment
+
     if experiment is None:
         # process all experiments
         experiments = client.list_experiments()
     else:
-        experiments = (client.get_experiment_by_name(experiment),)
+        if get_experiment_by_name:
```
alberttorosyan, line 51: lets use the single `--experiment` option.
try with name first, if experiment is not found try treat it as experiment ID
if the experiment is not found, echo an error message and exit

alberttorosyan (changes requested):

aim/cli/convert/processors/mlflow.py
devfox-se: well yeah good idea

alberttorosyan (approved):
please update description for `--experiment` argument for CLI ref. and guides.

Issue #1423, opened 2022-02-28
Chart interactions are not working in the Single Run Page Metrics tab


roubkar:
## 🐛 Bug

Chart interactions are not working in the Single Run Page Metrics tab due to the wrong `CSS Selecctor` being generated for the lines in `LineChart`

### To reproduce

- Track metrics with context that contains spaces
- Enter the Metrics tab of the Single Run Page
- Hover on `LineChart`

### Expected behavior

Active circle position and value should be updated when hovering on `LineChart`

### Environment

- Aim Version `v3.6.1`

PR #1424, opened 2022-02-28
[fix] Fix chart interactions issue in the Single Run Page Metrics tab

roubkar:
- Fix issue with wrong `CSS Selector` for lines in `LineChart`
- Sort metrics by name and context in Single Run Page


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1423

PR #1429, opened 2022-03-01
Cython Bindings / Setup

mahnerak:
This change configures all the extensions with third-party libs, shared objects and headers to build `aim` with.


alberttorosyan (approved):
lgtm

PR #1445, opened 2022-03-02
Fix text explorer url update

KaroMourad:
Text Explorer URL forward/backward doesn't work

- fix URL updating issue: missing config state from model initial state


rubenaprikyan (approved):
Lgtm!

PR #1453, opened 2022-03-03
docs/explorer update

ManeSah:
Started on line 134 finished on 200. To be continued, pls don't review yet


SGevorg:
The feature name should be `docs/explorer-update`

SGevorg (commented):
It seems the explorers docs structure (flat vs one explorers with sub paged) is under question.
Therefore this PR is pending before it gets fully agreed by @gorarakelyan @roubkar and @alberttorosyan

PR #1454, opened 2022-03-03
[fix] Fix ui rendering issue on colab

rubenaprikyan:
There is a javascript syntax error in time of rendering iframe. Also added a link which will open the ui in a new browser tab.


rubenaprikyan:
Resolves #1455

alberttorosyan (approved):
lgtm

PR #1460, opened 2022-03-04
[fix] Add record_density to initial get-batch request for figures

roubkar:
Change range and index setting from `RangePanel`


arsengit (approved):

mahnerak (approved):
lgtm

Issue #1360, opened 2022-02-17
Add Overview tab in single run page


arsengit:
## 🚀 Feature

Add Run Overview tab in single run page

### Motivation

Provide users nice ui to see overall and detailed info about run.  Also provide Metrics, Parameters, System Metrics, CLI Arguments, Environment Variables and Packages searchable tables.

### Additional context

![Screen Shot 2022-02-17 at 18 28 34](https://user-images.githubusercontent.com/44676488/154502782-57c6fe0d-97bc-44d8-9061-bf922042eb6c.png)
![Screen Shot 2022-02-17 at 18 29 23](https://user-images.githubusercontent.com/44676488/154502805-24e03aa7-6385-4616-8ee9-4c2d06c82335.png)
![Screen Shot 2022-02-17 at 18 36 51](https://user-images.githubusercontent.com/44676488/154503989-31f21536-7d55-4c6d-8fc3-291514f0178e.png)

PR #1477, opened 2022-03-08
[fix] Container padding issue in run single page tabs

arsengit:
- fixed container padding issue in single run page tabs


arsengit:
Resolves #1360

VkoHov (approved):
lgtm!

PR #1520, opened 2022-03-14
[feat] api profiler

devfox-se:


aim/cli/up/commands.py
```diff
@@ -39,7 +40,8 @@
                                                                 readable=True))
 @click.option('--base-path', required=False, default='', type=str)
 @click.option('--force-init', is_flag=True, default=False)
-def up(dev, host, port, workers, repo, tf_logs, ssl_keyfile, ssl_certfile, base_path, force_init):
+@click.option('--profiler', '-P', is_flag=True, default=False)
```
alberttorosyan, line 43: no need for a short flag. `--profile

aim/web/middlewares/profiler/__init__.py
```diff
@@ -0,0 +1 @@
+from ..profiler.profiler import PyInstrumentProfilerMiddleware
```
alberttorosyan, line 1: please use absolute imports

aim/web/middlewares/profiler/profiler.py
```diff
@@ -0,0 +1,81 @@
...
+            if message['type'] == 'http.response.start':
+                nonlocal status_code
+                status_code = message['status']
+            await send(message)
+
+        try:
+            await self.app(scope, receive, wrapped_send)
+        finally:
+            if scope["type"] == "http" and should_stop:
+                self._profiler.stop()
```
alberttorosyan, line 76: I think we can add a line covering request params and body.
Just need to keep HTML syntax.

alberttorosyan (changes requested):
left a couple of comments.
also agree with @mihran113 we should refer to the original repo with profiling middleware implementation

devfox-se:
Implements #1526

aim/web/middlewares/profiler/profiler.py
devfox-se: or we could just log it into a separate file with "almost" the same naming. Injecting into html is a bit challenging since its mostly JS code

aim/web/configs.py
```diff
@@ -7,3 +7,4 @@
 AIM_UI_TELEMETRY_KEY = 'AIM_UI_TELEMETRY_ENABLED'
 AIM_UI_BASE_PATH = '__AIM_UI_BASE_PATH__'
 AIM_PROXY_URL = '__AIM_PROXY_URL__'
+AIM_PROFILER_KEY = '__AIM_PROFILER_ENABLED'
```
alberttorosyan, line 10: pls add __ at the end as well

aim/web/middlewares/profiler/profiler.py
```diff
@@ -0,0 +1,102 @@
...
+                    "path": path,
+                    "body": body,
+                    "method": method,
+                    "params": params
+                }, separators=(',', ':'))
+
+                html_output = self._profiler.output_html(**self._profiler_kwargs)
+
+                # inject request data
+                html_output = html_output[:131] + f"<pre><code>{request_data}</code></pre>" + html_output[131:]
```
alberttorosyan, line 99: I guess this part is for injecting the `request_data` in `<body>`?
Is it safe to assume that generated html will be valid? the number of characters (131) might be different from version to version. Better to find the index of appropriate tag and use it instead of hard-coded value

alberttorosyan (changes requested):

aim/web/middlewares/profiler/__init__.py
alberttorosyan: there is still a relative import

aim/web/configs.py
devfox-se: how did i miss this 🤦‍♂️

aim/web/middlewares/profiler/profiler.py
devfox-se: agree

alberttorosyan (approved):
lgtm

PR #1532, opened 2022-03-16
[fix] Fix chart active point focusing issue

KaroMourad:
Expected behavior:
-  after focusing(clicking) on the active point, all charts should freeze(block mouse hovering on charts) before you unfocus the active point, it means before you don't click on one of the sibling charts 

Bug: 
- after focusing on the active point, it doesn’t focus

Solution: 
- on syncing hover state for charts, we immediately set the focused state for every sibling charts and itself

Improvement: 
- Improve d3 utils `drawHoverAttributes` function performance
    1. `drawAxisLabel` (create/remove/create changed to create/update)
    2. `drawHoverLine` (create/remove/create changed to create/update)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixes:
 
+- Fix issue with focusing on the chart active point while moving the cursor (KaroMourad)
 - Fix the image full view toggle icon visibility if the image has a white background (VkoHov)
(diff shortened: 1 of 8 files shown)


KaroMourad:
Resolves #1534

VkoHov (approved):
lgtm!

PR #1539, opened 2022-03-18
[fix] refactored xgboost callback

devfox-se:


aim/sdk/adapters/xgboost.py
```diff
@@ -3,70 +3,59 @@
...
+try:
+    from xgboost.callback import TrainingCallback
+except ImportError:
+    raise RuntimeError(
+        'This contrib module requires XGBoost to be installed. '
+        'Please install it with command: \n pip install xgboost'
+    )
+
+
+class AimCallback(TrainingCallback):
```
mihran113, line 15: Looks good, but it would be nice to have `experiment` property for an access to underlying `Run` instance, as it's done here:
https://github.com/aimhubio/aim/blob/4f087862d0720493ffc83a5916cd97a76d913959/aim/sdk/adapters/pytorch_lightning.py#L49

devfox-se:
Covers #1542

alberttorosyan (commented):
lgtm.
@devfox-se can you please make sure that the existing docs are covering the basic logger integration? Please refer to PL or Pytorch Ignite as a reference (just try to find all references to PL/PI in the docs)

mihran113 (approved):
lgtm

Issue #1538, opened 2022-03-18
PyTorch Lightning adapter fails to log hyperparams when a git repo is initialized


RomDeffayet:
## 🐛 Bug

Starting from 3.6, Pytorch Lightning adapter fails to log hyper-parameters when a git repo is initialized

### To reproduce

1 - `git init`
2 - Run any PL script with Aim logging hyperparameters. 
<details>
<summary>For example</summary>

```
import os
import torch
from torch import nn
import torch.nn.functional as F
from torchvision import transforms
from torchvision.datasets import MNIST
from torch.utils.data import DataLoader, random_split
import pytorch_lightning as pl
from aim.pytorch_lightning import AimLogger
from aim import Run, Figure
from typing import Optional, Any
from pytorch_lightning.callbacks import RichProgressBar, ModelCheckpoint

class MNISTDataModule(pl.LightningDataModule):
    def __init__(self, data_dir: str = "./"):
        super().__init__()
        self.data_dir = data_dir
        self.transform = transforms.Compose([transforms.ToTensor(), transforms.Normalize((0.1307,), (0.3081,))])

        # Setting default dims here because we know them.
        # Could optionally be assigned dynamically in dm.setup()
        self.dims = (1, 28, 28)

    def prepare_data(self):
        # download
        MNIST(self.data_dir, train=True, download=True)
        MNIST(self.data_dir, train=False, download=True)

    def setup(self, stage: Optional[str] = None):

        # Assign train/val datasets for use in dataloaders
        if stage == "fit" or stage is None:
            mnist_full = MNIST(self.data_dir, train=True, transform=self.transform)
            self.mnist_train, self.mnist_val = random_split(mnist_full, [55000, 5000])

            # Optionally...
            # self.dims = tuple(self.mnist_train[0][0].shape)

        # Assign test dataset for use in dataloader(s)
        if stage == "test" or stage is None:
            self.mnist_test = MNIST(self.data_dir, train=False, transform=self.transform)

            # Optionally...
            # self.dims = tuple(self.mnist_test[0][0].shape)

    def train_dataloader(self):
        return DataLoader(self.mnist_train, batch_size=32, num_workers = 32)

    def val_dataloader(self):
        return DataLoader(self.mnist_val, batch_size=32, num_workers = 32)

    def test_dataloader(self):
        return DataLoader(self.mnist_test, batch_size=32, num_workers = 32)

class LitAutoEncoder(pl.LightningModule):
    def __init__(self):
[...]

PR #1540, opened 2022-03-18
[fix] git info collection

devfox-se:
Added more checks when collecting git info. 
In case if one command fails its more likely others will fail too. In such case no data is collected and the process is stopped.

### DIFF
--- a/aim/ext/utils.py
+++ b/aim/ext/utils.py
@@ -29,46 +29,45 @@ def get_git_info():
                            stderr=subprocess.STDOUT,
                            check=True)
     except subprocess.CalledProcessError:
-        git_active = False
+        # not a git repo
+        return git_info
     else:
-        output = r.stdout.decode('utf-8').strip()
-        if output == 'true':
-            git_active = True
-        else:
-            git_active = False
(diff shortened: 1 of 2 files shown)


devfox-se:
Covers #1538

aim/ext/utils.py
```diff
@@ -23,23 +23,42 @@ def get_environment_variables():
...
+        ('git', 'rev-parse', '--abbrev-ref', 'HEAD'),
+        ('git', 'config', '--get', 'remote.origin.url'),
+        ('git', 'log', '--pretty=format:%h/%ad/%an', '--date=iso-strict', '-1'),
+    )
+    results = []
+    for cmd in cmds:
+        try:
+            r = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, check=True)
+        except subprocess.CalledProcessError:
+            print('Something went wrong while collecting git info. Will skip this step.')
```
mihran113, line 50: @devfox-se let's add a logger instead of basic print

alberttorosyan (approved):
lgtm

mihran113 (approved):
lgtm

Issue #1546, opened 2022-03-20
Incorrect calculation of active point on mouse hover in LineChart


KaroMourad:
## 🐛 Bug

The active point should be calculated depending on the cursor's closest x-axis value, which is calculated incorrectly

### Expected behavior

Expected to have active point on mouse hover, which is calculated depending on the cursor's closest x-axis value and the nearest point by radius

### Environment

- Aim Version (v3.7.4)

### Additional context

https://user-images.githubusercontent.com/44981312/159174106-0c68ec66-b8e3-4c8c-8352-3412ade45c08.mov

PR #1547, opened 2022-03-20
[fix] Fix issue with calculation of active point on mouse hover in the LineChart

KaroMourad:
fixed  `getClosestCircle` function in d3 utils `drawHoverAttributes` function


KaroMourad:
Resolves #1546

rubenaprikyan (approved):
looks good

arsengit (approved):

PR #1554, opened 2022-03-22
[fix] API project repo container pools caching

mihran113:
Cleanup repo container pools for each request so in progress run containers won't be cached. 
This is causing in-progress runs to show the same result in audios, images and etc tabs.


mihran113:
Resolves #1557

alberttorosyan (approved):
lgtm

PR #1563, opened 2022-03-23
[fix] Retry docker image build when PyPI package is newly pushed

alberttorosyan:
Added retry logic to docker image creation bash script.  Retry 5 times with 2 min. interval. Expected to pass after 2nd attempt, the rest are kept just in case


alberttorosyan:
resolves #1564

mihran113 (approved):
lgtm

PR #1565, opened 2022-03-23
[feat] get step of sequence api

devfox-se:


devfox-se:
Implements #1566

aim/web/api/runs/object_views.py
```diff
@@ -95,6 +95,38 @@ async def sequence_batch_api(run_id: str,
...
+            api = CustomObjectApi(seq_name, resolve_blobs=cls.resolve_blobs)
+            api.set_dump_data_fn(cls.dump_record_fn)
+            api.set_requested_traces(run, requested_traces)
+
+            if record_step is None:
+                total_record_range = api.get_total_record_range()
+                if api.use_list:
+                    total_record_range = total_record_range[0]
+                record_step = total_record_range.stop - 1
+            record_range = checked_range(f"{record_step}:{record_step + 1}")
```
alberttorosyan, line 124: why convert to string then back to tuple?
`record_range = IndexRange(record_step, record_step + 1)`

aim/web/api/runs/object_views.py
```diff
@@ -95,6 +95,38 @@ async def sequence_batch_api(run_id: str,
...
+            run = repo.get_run(run_id)
+            if not run:
+                raise HTTPException(status_code=404)
+
+            api = CustomObjectApi(seq_name, resolve_blobs=cls.resolve_blobs)
+            api.set_dump_data_fn(cls.dump_record_fn)
+            api.set_requested_traces(run, requested_traces)
+
+            if record_step is None:
+                total_record_range = api.get_total_record_range()
```
alberttorosyan, line 120: total range will be calculated twice. 1st time here, 2nd time after `set_ranges` call. better to leave this logic in `CustomObjectApi` and do the calculation the only once

alberttorosyan (changes requested):
left some comments

aim/web/api/runs/object_api_utils.py
```diff
@@ -86,6 +90,11 @@ def set_ranges(self, record_range: IndexRange, record_density: int,
         else:
             self.total_record_range = self._calculate_ranges()
 
+        if record_step is not None:
+            if record_step < 0:
+                record_step += self.total_record_range.stop
```
alberttorosyan, line 95: do we have a real use-case for this? if so, still need to check that modulo of record_step is in the allowed range.

alberttorosyan: the alternative to specifying -1 as a default would be `None` and checking that ether `record_range` or `record_step` is not None

devfox-se: yeah that was the initial method, yet i think its confusing

devfox-se: I added assertion just in case

alberttorosyan (approved):

PR #1577, opened 2022-03-25
[fix] profiler issues

devfox-se:
- hangs on `get-batch`
- not working on `<4.0.0`

### DIFF
--- a/aim/web/middlewares/profiler/profiler.py
+++ b/aim/web/middlewares/profiler/profiler.py
@@ -13,6 +13,15 @@ from starlette.routing import Router
 from starlette.requests import Request
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
+try:
+    from pyinstrument import Profiler, __version__
+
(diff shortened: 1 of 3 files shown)


mihran113:
@devfox-se I think destination branch needs to be changed to `release/3.8.x`

aim/web/middlewares/profiler/profiler.py
```diff
@@ -13,6 +13,15 @@
 from starlette.requests import Request
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
+try:
```
alberttorosyan, line 16: lets still keep this at class level. otherwise pyinstrument will be imported even without running with --profile flag

devfox-se: then we have to re-import this on every api call

devfox-se: disregard

alberttorosyan (approved):
lgtm

PR #1578, opened 2022-03-25
[fix] Fix issue with rendering lines for large range scale

KaroMourad:
- when ignoring outliers, crop lines to avoid rendering large range scale
- same cropping calculations used for aggregated data: area, line

- add filtering `epochs`, `timestamps` invalid data in `filterMetricData` function


rubenaprikyan (approved):
Lgtm

roubkar (approved):

PR #1600, opened 2022-03-29
[fix] Fix incorrect chart active point values for large range scale

KaroMourad:
After inverting the active point mouse position we get an `x` and `y` value, they can be different from real values for large range scale.
To fix the issue we get the closest actual value by using the `d3.bisectCenter` function.


KaroMourad:
Resolves #1602

rubenaprikyan (approved):
lgtm!

roubkar (approved):

Issue #1605, opened 2022-03-31
[Bug] Runs table not loading after 2-3 pages of results


ashutoshsaboo:
## 🐛 Bug

Runs table not loading after 2-3 pages of results (45 runs in each pagination call).

### To reproduce

Described above log large number of runs try to scroll down on runs table page.

### Expected behavior

All runs should load as expected.

rubenaprikyan:
@ashutoshsaboo is the fix validated ? let's close this issue if yes.

ashutoshsaboo:
@rubenaprikyan yes opens up fine for now, will re-open and report if I see a similar issue again. Thanks so much for a prompt fix! :D

PR #1606, opened 2022-03-31
[fix] Fix issue with fetch on load more (infinite scroll) functionality in Runs Explorer

rubenaprikyan:
There was a logic to update the request limit on state for live update, live update should get all data, but it shouldn't update the limit exactly on state, because the infinite scroll uses that limit, and it should always have constant value.
Now the I added a calculation logic for live update, and using it once the live updates' start function is called.

Please do your tests with live update enabled status and without it as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.8.1
+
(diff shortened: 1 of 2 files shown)


rubenaprikyan:
Resolves #1605

roubkar (approved):

VkoHov (approved):

Issue #1608, opened 2022-03-31
`RepoIndexManager` keeps run references


mihran113:
## 🐛 Bug

Not properly finalized runs will be locked indefinitely when using aim UI, as `RepoIndexManager` keeps reference on the `aim.Run` instance.

### To reproduce

```
from aim import Run
import time
run = Run()
run['foo'] = 'bar'
run.track(1, 'test_metric')
time.sleep(20)
```
Kill the script with `SIGKILL` signal, so the run won't be finalized properly, then open UI with `aim up` command and then try to delete the run:
```
Error while trying to delete run '6c7e0746cd954d49b541f7ab'. The file lock '/Users/mvanyan/playground/remote_test/.aim/meta/locks/6c7e0746cd954d49b541f7ab' could not be acquired..
INFO:     127.0.0.1:61853 - "DELETE /api/runs/6c7e0746cd954d49b541f7ab HTTP/1.1" 400 Bad Request

```
### Expected behavior

Finalized runs should be deletable (not locked) from aim UI

### Environment

- Aim Version: 3.8.0
- Python version: 3.6 
- pip version: 21.3.1
- OS: MacOS

PR #1609, opened 2022-03-31
[fix] Run reference cleanup from `RepoIndexManager` 

mihran113:
- Remove run reference from `RepoIndexManager` thread after explicitly finalizing the run instance
- Pass the repo path to index manager instead of repo instance to avoid infinitely alive reference on repo instance


mihran113:
Resolves #1608

alberttorosyan (approved):
lgtm

Issue #1610, opened 2022-03-31
Runs are not finalized when using remote tracking server


mihran113:
## 🐛 Bug

When using remote repos to track metrics, runs are not finalized after script finishes successfully.

### To reproduce
Start remote tracking server with  `aim server` command then run this script

```
from aim import Run
run = Run(repo='')
run['foo'] = 'bar'
run.track(1, 'test_metric')

```

Then run the command below
```
ls .aim/meta/progress
```

This will show that there's a run in progress which should not been the case, as the script was properly finished.

### Expected behavior

Runs should not stay in `in progress` state if the script wasn't force killed.

### Environment

- Aim Version: 3.8.0
- Python version: 3.6
- pip version: 21.3.1
- OS: MacOS

PR #1612, opened 2022-03-31
[fix] Remote run finalization

mihran113:
- Provide `.finalize()` method for `SubTreeView` class
- Make `finalize()` read instruction instead of write instruction
- Remove implicit RpcQueue cleanup and leave it to be handled by Run resource cleanup


alberttorosyan (approved):
lgtm
lets wait for @aramaim's feedback as well

mihran113:
Resolves #1610

Issue #1613, opened 2022-04-01
Row visualization functionality is broken


roubkar:
## 🐛 Bug

`Hide Row` functionality removes/hides the row instead of attaching an `inactive` state and only removes the corresponding line from the chart

### To reproduce

- Go to Metrics Explorer
- Click on the `eye` icon on the table last column

### Expected behavior

The corresponding row should be displaying with an inactive state

### Environment

- Aim Version 3.8.1

PR #1614, opened 2022-04-01
[fix] Fix row hiding functionality

roubkar:
- Apply `inactive` class instead of `hidden` on rows that are not visualized
- Fix layout/content shifts for chart panel and table when switching from request pending state


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1613

arsengit (approved):

VkoHov (approved):
lgtm!

Issue #1607, opened 2022-03-31
Implement data model and api interface for Note feature on BE


devfox-se:
## 🚀 Feature

Add BE support for Run notes which will help users to quickly save and update notes about their runs

### Motivation

This feature will help users keep one or more extra notes about their results which they can reference to later.

PR #1620, opened 2022-04-03
[feat] Notes BE implementation

devfox-se:


devfox-se:
Implements #1607

aim/storage/structured/sql_engine/models.py
```diff
@@ -85,3 +93,45 @@ def __init__(self, name):
...
+
+
+class Note(Base):
+    __tablename__ = 'note'
+    __table_args__ = (
+        UniqueConstraint('run_id', 'name'),
+    )
+
+    id = Column(Integer, autoincrement=True, primary_key=True)
+    name = Column(Text, nullable=False, default='New Note')
```
alberttorosyan, line 105: what the `name` field is used for?

aim/storage/structured/sql_engine/models.py
```diff
@@ -85,3 +93,45 @@ def __init__(self, name):
     def validate_color(self, _, color):
         # TODO: [AT] add color validation
         return color
+
+
+class Note(Base):
+    __tablename__ = 'note'
+    __table_args__ = (
```
alberttorosyan, line 100: do we need this constraint? why restrict someone to add the same note (comment) multiple times?

aim/web/api/runs/note/pydnamic_models.py
```diff
@@ -0,0 +1,6 @@
+from pydantic import BaseModel
```
alberttorosyan, line 1: please move the APIs to `runs/views.py`
the directory structure becomes too nested here.
same for the inout models

aim/storage/structured/sql_engine/entities.py
```diff
@@ -175,6 +180,78 @@ def remove_tag(self, tag_name: str) -> bool:
         session.flush()
         return tag_removed
 
+    @property
+    def notes(self) -> List[dict]:
```
alberttorosyan, line 184: worth adding `notes_obj` method like it is done for tags 🤔

alberttorosyan (changes requested):
left some comments.
please add the unit-tests as well

aim/storage/structured/sql_engine/models.py
devfox-se: @alberttorosyan treat is as a file name where you write your notes

aim/storage/structured/sql_engine/entities.py
devfox-se: maybe... but what would it give us?

aim/storage/structured/sql_engine/models.py
devfox-se: @alberttorosyan its like having `new folder` and `new folder` on your desktop :D how would you distinguish ?

alberttorosyan (approved):
lgtm

Issue #1625, opened 2022-04-05
CSS valid selector issue with Uncaught DOMException


ptaejoon:
## 🐛 Bug

if run_hash has improper name concerned with css selector, some UI function does not work.
Error Message :
Table.tsx:454 Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Document': '.Table__cell.rowKey-sample_metric_dag-scheduled__2022-04-03T15-10-00+00-00' is not a valid selector.
Since + is not valid character for css selector, the error is errupted

I found this problem stops hovering on graph in metrics tab.
### To reproduce
aim.Run(run_hash="improper_name_which_out_of_[azAZ09_-]" ... ) (ex. aim.run(run_hash="31") -- id start with number is not valid css selector.) (https://www.geeksforgeeks.org/which-characters-are-valid-in-css-class-names-selectors/)

### Expected behavior
This issue is caused by naming css element following run_hash. it will not happen if aim.Run is declared without setting run_hash. Writing this issue on guide or changing name properly inside package can be useful.

### Environment

- Aim Version (e.g., 3.0.1) 3.6
- Python version 3.8.2
- pip version 22.0.4
- OS (e.g., Linux) mac
- Any other relevant information

RamiSJ12:
I'm assuming clickField is supposed to be the class of the element you want to get. In that case, since the class starts with a digit, the easiest way to select it using CSS selector is using attribute selectors, like so:

document.querySelector(`[class="${clickField}"]`).innerHTML = `<img src ='${icon}.svg'>`

If clickField might contain " characters, you need to replace ${clickField} by ${clickField.replace(/"/g, '\\"')} to escape them properly.

There is also a far easier way, using document.getElementsByClassName():

document.getElementsByClassName(clickField)[0].innerHTML = `<img src ='${icon}.svg'>`
Classes may contain any characters except space, as the [HTML standard](https://html.spec.whatwg.org/multipage/dom.html#classes) imposes no restrictions on them.

roubkar:
Hey @ptaejoon, thanks for reporting.
Looking into it.

Hamik25:
Hey @ptaejoon, the fix for this issue has been shipped with v3.8.1
To upgrade please run pip install aim --upgrade

PR #1628, opened 2022-04-05
[fix] Encode run hash before including in CSS selectors

Hamik25:
- Add encoding to key property whose value contains `run.hash` for explorers `processData` functions to avoid setting invalid CSS selectors on HTML tags. 
- Add encoding to selectors `run.hash` for metric lines.


Hamik25:
Resolves #1625

VkoHov (approved):

roubkar (approved):

Issue #1619, opened 2022-04-03
AimLogger error on new Pytorch-Lightning Release


haritsahm:
## 🐛 Bug

Removed `_convert_params` function in new pytorch-lightning release (1.6.0). 

### To reproduce

```
from aim.pytorch_lightning import AimLogger

trainer = pl.Trainer(logger=AimLogger(experiment='experiment_name'))
trainer.logger.log_hyperparams(hparams)
```
### Expected behavior

- Successfully log the hyperparameters with no errors

### Environment

- Aim Version (3.8.0)
- Python version (3.8.12)
- pip version (21.2.4)
- OS (e.g., Linux)
- Pytorch-lightning (1.6.0)
## Error output
```
Error executing job with overrides: ['experiment=dma_net', 'debug=default', 'log_dir=debug', 'data_dir=/media/haritsahm/DataStorage/dataset/cityscapes/cityscape_fo_segmentation/', 'logger=aim']
Traceback (most recent call last):
  File "train.py", line 22, in main
    return train(config)
  File "/home/haritsahm/Documents/projects/deeplearning/pytorch-DMANet/src/training_pipeline.py", line 74, in train
    utils.log_hyperparameters(
  File "/opt/conda/lib/python3.8/site-packages/pytorch_lightning/utilities/rank_zero.py", line 32, in wrapped_fn
    return fn(*args, **kwargs)
  File "/home/haritsahm/Documents/projects/deeplearning/pytorch-DMANet/src/utils/__init__.py", line 143, in log_hyperparameters
    trainer.logger.log_hyperparams(hparams)
  File "/opt/conda/lib/python3.8/site-packages/pytorch_lightning/utilities/rank_zero.py", line 32, in wrapped_fn
    return fn(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/aim/sdk/adapters/pytorch_lightning.py", line 70, in log_hyperparams
    params = self._convert_params(params)
AttributeError: 'AimLogger' object has no attribute '_convert_params'

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
root@0e330cd4b21f:/home/haritsahm/Documents/projects/deeplearning/pytorch-DMANet# 

```

mihran113:
[...]

PR #1637, opened 2022-04-08
Fix compatibility with PL v1.6.0

mihran113:
Have a copy of removed staticmethod `_convert_params()` on `AimLogger` class itself


mihran113:
Resolves #1619

alberttorosyan (approved):
lgtm

PR #1641, opened 2022-04-08
[fix] `PATH` related issues 

mihran113:
Call alembic and uvicorn via python and not directly to be sure that executables are present.

### DIFF
--- a/aim/cli/up/utils.py
+++ b/aim/cli/up/utils.py
@@ -17,8 +17,8 @@ def build_db_upgrade_command():
 
 def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile, log_level):
     cmd = [sys.executable, '-m', 'uvicorn',
-           '--host', host, '--port', '%s' % port,
-           '--workers', '%s' % num_workers]
+           '--host', host, '--port', f'{port}',
+           '--workers', f'{num_workers}']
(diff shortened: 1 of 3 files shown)


devfox-se (approved):
very good approach

aim/cli/up/utils.py
```diff
@@ -11,11 +12,13 @@ def build_db_upgrade_command():
...
         ini_file = os.path.join(migrations_dir, 'alembic_dev.ini')
-    return ['alembic', '-c', ini_file, 'upgrade', 'head']
+    return [sys.executable, '-m', 'alembic', '-c', ini_file, 'upgrade', 'head']
 
 
 def build_uvicorn_command(host, port, num_workers, ssl_keyfile, ssl_certfile, log_level):
-    cmd = ['uvicorn', '--host', host, '--port', '%s' % port, '--workers', '%s' % num_workers]
+    cmd = [sys.executable, '-m', 'uvicorn',
+           '--host', host, '--port', '%s' % port,
+           '--workers', '%s' % num_workers]
```
devfox-se, line 21: `'%s'`  - this is just completely obsolete

mihran113: it actually is not, as `subprocess` requires all the list members to be `str` values

alberttorosyan: replace with f-string?
`f'num_workers'`

alberttorosyan (approved):

Issue #1633, opened 2022-04-06
Images haven't been rendered in their original size


VkoHov:
## 🐛 Bug

Images have been cropped and haven't been rendered in their original size

### To reproduce
1. Go to Images Explorer
2. Align images by `Original Size`

<img width="966" alt="Screen Shot 2022-01-25 at 17 56 41" src="https://user-images.githubusercontent.com/43929589/162045852-424bc581-6f1c-436a-9d95-00cc6503eb20.png">

### Expected behavior

The rendered size has to match the intrinsic size

### Environment

- Aim Version v3.8.0

PR #1646, opened 2022-04-10
[fix] The image's original size cropping

VkoHov:
Fix the rendering of images aligned by the original size


VkoHov:
Resolves [#1633](https://github.com/aimhubio/aim/issues/1633)

arsengit (changes requested):
@VkoHov The `SPACE_BETWEEN_ITEMS` value should be the same with the gap between the items

arsengit (approved):

rubenaprikyan (approved):
lgtm!

PR #1651, opened 2022-04-13
[feat] Improve chart lines rendering performance

KaroMourad:
- use `shape-rendering: optimizeSpeed` CSS property in HighPlot and LineChart components when the display is small with high `dpi` and  the lines count is more than `limit`
- replace meaningless `stroke-dasharray: '5 0'` to `stroke-dasharray: 'none'` which is default value to improve rendering

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Enhancements:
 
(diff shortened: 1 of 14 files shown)


arsengit (approved):

VkoHov (approved):
lgtm!

gorarakelyan:
resolves #1655

PR #1658, opened 2022-04-14
[feat] Optimize stream decoding performance on UI

mahnerak:
Reduce the number of microtasks made for each chunk processed.
With this change the number of async calls is not elevated as every chunk yielded from `fetch` is maximally processed in synchronous/blocking manner.

The chunk size is auto-adjusted by `fetch`. While testing I haven't noticed any UI blocking / lag yet.


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1679

roubkar (approved):

aim/web/ui/src/utils/encoder/streamEncoding.ts
```diff
@@ -101,12 +101,12 @@ function decodeByType(typeId: number, buffer: ArrayBuffer) {
   return value;
 }
 
-function typedArrayToBuffer(array: Uint8Array): ArrayBuffer {
-  return array.buffer.slice(
-    array.byteOffset,
-    array.byteLength + array.byteOffset,
-  );
-}
+// function typedArrayToBuffer(array: Uint8Array): ArrayBuffer {
```
mahnerak, line 104: @roubkar do we want it to stay as comment?

Issue #1661, opened 2022-04-14
Add LightGBM integration


devfox-se:
## 🚀 Feature

We want to implement a callback for LightGBM which will hook and track data processed by the package.

Example of callbacks: https://lightgbm.readthedocs.io/en/latest/Python-API.html#callbacks

### Motivation

Integration with LightGBM will make aim extend its usecases and let users get track their data trough Aim

PR #1662, opened 2022-04-14
[feat] Add AimCallback for LightGBM

devfox-se:


devfox-se:
Implements #1661

aim/sdk/adapters/lightgbm.py
```diff
@@ -0,0 +1,97 @@
+from typing import Optional
+
+from aim import Run
+from aim.ext.resource import DEFAULT_SYSTEM_TRACKING_INT
+
+try:
+    from lightgbm.callback import CallbackEnv
+except ImportError:
+    raise RuntimeError(
```
mahnerak, line 9: This one will always warn (even if you don't use LightGBM adapter) doesn't it?

alberttorosyan: @mahnerak error will be raised only if you try to import the adapter. as far as we don't have any imports in `aim` or `aim.sdk` modules, we should be fine.
@devfox-se please confirm

aim/sdk/adapters/lightgbm.py
```diff
@@ -0,0 +1,97 @@
...
+                 log_system_params: Optional[int] = True):
+        self._repo_path = repo
+        self._experiment = experiment
+        self._system_tracking_interval = system_tracking_interval
+        self._log_system_params = log_system_params
+        self._run = None
+        self._run_hash = None
+
+        # callback parameters
+        self.order = 15
```
alberttorosyan, line 35: what `order` stands for?

aim/sdk/adapters/lightgbm.py
```diff
@@ -0,0 +1,97 @@
...
+                data_name, eval_name, result, _ = item
+                self._run.track(result, name=eval_name, context={'data_name': data_name})
+            else:
+                data_name, eval_name = item[1].split()
+                res_mean = item[2]
+                res_stdv = item[4]
+                self._run.track(res_mean, name=f'{eval_name}-mean', context={'data_name': data_name})
+                self._run.track(res_stdv, name=f'{eval_name}-stdv', context={'data_name': data_name})
+
+        self.before_tracking(env=env)
```
alberttorosyan, line 88: should be `after_tracking` I guess

docs/source/using/integration_guides.md
```diff
@@ -184,3 +184,22 @@ class CustomCallback(AimCallback):
...
+
+class CustomCallback(AimCallback):
+
+    def before_tracking(self, env):
+        for item in env.evaluation_result_list:
+            # manipulate item here
+            pass
+
+    def after_tracking(self, env):
+        self.experiment.track(env.iteration, name="Iteration")
```
alberttorosyan, line 204: is this the iteration step? what about adding some actual metric or maybe other objects?

alberttorosyan (changes requested):
left a couple of comments; overall looks good!
I think we can also add the LigtGBM logo in docs landing page and GH readme 🤔

aim/sdk/adapters/lightgbm.py
devfox-se: good catch

aim/sdk/adapters/lightgbm.py
devfox-se: no there is no problem with it, we have this technique used in all of our callbacks

aim/sdk/adapters/lightgbm.py
devfox-se: I cannot really say, its not documented and all the `LightGBM`'s callbacks have order set in them, I picked a value in between

devfox-se:
@alberttorosyan Regarding the landing page logo - we need something prepared for it to align with others

docs/source/using/integration_guides.md
devfox-se: yeah its really hard to come up with something more useful 😆

aim/sdk/adapters/lightgbm.py
alberttorosyan: @mahnerak any idea here?

mahnerak: When there are multiple Callbacks registered this defines the precedence of their calls.
I think this value should be as great (as late) as possible, but before checkpointing.

It looks like `30` is for early stopping and serialization of the model, and `20` for recording some logs. So something in between will work better.

alberttorosyan: @devfox-se please double-check this.

alberttorosyan (approved):

Issue #1665, opened 2022-04-14
Add documentation for the info massage in the slider of Range Panel  which has only 1 value tracked


VkoHov:
## 📚 Documentation

Add documentation for the info message in the slider of the `Range Panel`  component which has only 1 value tracked.

### Motivation

Provide users with better knowledge about the `Range Panel` component.

PR #1666, opened 2022-04-14
[docs] Added range panel info message docs

VkoHov:
Add documentation for the info message in the slider of the Range Panel which has only 1 value tracked


VkoHov:
Resolves [#1665](https://github.com/aimhubio/aim/issues/1665)

roubkar (approved):

PR #1669, opened 2022-04-15
[fix] Notes tab dynamic import chunk name

arsengit:
- Fixed Notes tab dynamic import chunk name
- Changed `font-display` CSS property  from swap to block in font-faces
- Removed unused imported modules


VkoHov (approved):
lgtm!

PR #1677, opened 2022-04-19
[feat] Add AimLogger for CatBoost

devfox-se:


devfox-se:
Implements #1680

aim/sdk/adapters/catboost.py
```diff
@@ -0,0 +1,94 @@
...
+    @property
+    def experiment(self) -> Run:
+        if not self._run:
+            self.setup()
+        return self._run
+
+    def setup(self):
+        if self._run:
+            return
+        if self._run_hash:
```
alberttorosyan, line 39: do we need the re-opening logic here? as I understand, the callback is used as a logger (much like the  stdout) so normally, it should be just deleted once the process is finished.

aim/sdk/adapters/catboost.py
```diff
@@ -0,0 +1,94 @@
+from sys import stdout
+from typing import Optional
+
+from aim import Run
+from aim.ext.resource import DEFAULT_SYSTEM_TRACKING_INT
+
+
+class AimCallback:
```
alberttorosyan, line 8: lets rename to `AimLogger`

alberttorosyan (changes requested):
looks great! a couple of comments left

aim/sdk/adapters/catboost.py
devfox-se: leaving it wont have any negative impact I assume

alberttorosyan (approved):

PR #1692, opened 2022-04-22
[feat] Add step to validate PR naming convention

devfox-se:
The format of the message should follow the template below

```
[tag] Subject line
```

This PR enforces the following rules

🟢 Capitalize the first letter of the subject.
🟢 Do not end with subject line with dot (.)
🔴 Start the subject line with verb and use the imperative mood.
🟢 Tags are: [feat] [fix] [doc] [refactor]
🚫 When merging PR you might need to adjust auto-generated commit message (GH will put the PR number in the end).

We decided to retain PR number suffix (e.g. `(#1692)`) upon merge


alberttorosyan (commented):
looks good!
why can't we merge this with the `pull-request` action workflow? I see that events are different, but we can add if statement to the steps and run only those which are required

devfox-se:
@alberttorosyan i see no reason to complicate stuff:)

alberttorosyan:
> @alberttorosyan i see no reason to complicate stuff:)

why you think this complicates things? keeping workflows tight is better than having multiple workflows for same events.

devfox-se:
And here goes the first naming convention enforced merge ;)

PR #1693, opened 2022-04-22
[fix] Params/Scatters explorers' row hiding functionality

VkoHov:
- Replaced run hush with encoded key.
- Remove needless `pointer-events: all` from the area drawing method.


KaroMourad (approved):

VkoHov:
Resolves #1694

rubenaprikyan (approved):
lgtm!

PR #1695, opened 2022-04-22
[fix] HF callback context capturing 

mihran113:
Added also step/epoch values to tracked metrics and process rank 0 checks


alberttorosyan (approved):

mihran113:
Resolves #1724

PR #1703, opened 2022-04-26
Fix incorrect type hints in Repo class.

### LINKED ISSUE #1702: Pylance reports code is unreachable
## 🐛 Bug

**image**

Pylance reports code is unreachable after creating an `Repo` instance. I believe this is due to the incorrect type hints in the `Repo` class.

Pylance sees `read_only` is a `bool`, so the error is always raised and the instantiation is believed to be never successful.

uduse:
Fix #1702

### DIFF
--- a/aim/sdk/repo.py
+++ b/aim/sdk/repo.py
@@ -92,7 +92,7 @@ class Repo:
 
     tracking_queue = _get_tracking_queue()
 
-    def __init__(self, path: str, *, read_only: bool = None, init: bool = False):
+    def __init__(self, path: str, *, read_only: Optional[bool] = None, init: Optional[bool] = False):
         if read_only is not None:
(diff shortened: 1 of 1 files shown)


alberttorosyan (changes requested):
@uduse thanks for contribution!
may I ask to list the fix in CHANGELOG.md as well?

uduse:
@alberttorosyan Sorry I'm not sure what's the best way for me to add this change. I used GitHub's edit function and it created the patch for me. Should I have a local copy of the repo or something?

alberttorosyan:
@uduse you can just add changes on top of `patch-1` branch of the forked repository. Once you push it to remote, the PR will automatically include the CHANGELOG updates 🙌

alberttorosyan (approved):

PR #1710, opened 2022-04-28
[fix] LineChart y-dimension margin calculation for negative values

KaroMourad:
Fix: 
- margin calculation for negative and small values [0, 1]
- smoothing algorithm when empty `y` and `x` values


KaroMourad:
Resolves #1711

rubenaprikyan (approved):
lgtm!

Issue #1698, opened 2022-04-25
Keep the state of the chart while re-rendering


VkoHov:
## 🚀 Feature

Persist states inside charts (LineChart, HighPlot, ScatterPlot): apply current modifications while re-rendering the chart.

### Motivation

To avoid the extra actions for configuration of the chart and also avoid losing an already configured state.

### Pitch

1. Decrease the count of the chart re-rendering 
2. Keep the local state of the chart until it will finish re-rendering

PR #1715, opened 2022-05-01
[feat] Keep the local state of the charts

VkoHov:
- Keep the Highplot axis brash position in the model state
- Restore the Highplot axis brush position from the state
- Hide chart tooltip content double scrollbar 
- Reset brush extents from the model state after grouping apply
- Update brush size after window resize


VkoHov:
Resolves #1698

KaroMourad (approved):

Issue #1720, opened 2022-05-03
Changing the color persistence of the charts affects stroke types


KaroMourad:
## 🐛 Bug

Changing the color persistence of the charts affects stroke types

### To reproduce

1. Apply grouping by stroke and color
2. Enable stroke and color persistence
3. Shuffle color persistence
4. Explore changing stroke style type

### Expected behavior

Expected to not change stroke types after changing the color persistence of the charts

### Environment

- Aim Version (v3.9.2)

PR #1721, opened 2022-05-03
[fix] Fix affecting stroke types after changing color persistence

KaroMourad:
fix: 
- replace the hardcoded `color` variable with dynamic


KaroMourad:
Resolves #1720

Issue #1723, opened 2022-05-03
Lines rendered partially in the single run metrics and systems tabs


KaroMourad:
## 🐛 Bug

When rendering metrics and system-metrics charts in the single run details `metrics` and `system` tabs, depending on logs, the chart lines rendered partially

### To reproduce

1. open single run details `metrics` tab
2. explore rendering lines in LineChart

For reproduction, you need to have metric logs with `falsy` values

### Expected behavior

Expected to draw metrics correctly without cutting lines 

### Environment

- Aim Version (v3.9.2)

### Additional context

![metrics](https://user-images.githubusercontent.com/44981312/166518009-4e1282d8-4bf5-4e0d-938e-2f6bf11779a9.png)

PR #1726, opened 2022-05-04
[fix] Fix issue with filtering metrics values in single run page

KaroMourad:
Add:
- `filterSingleRunMetricsData` util function

Fix:
- filter single run metrics data
- `filterMetricsData` util function


KaroMourad:
Resolves #1723

roubkar (approved):

Issue #1739, opened 2022-05-11
Move user-selected params to the top in chart popover


roubkar:
## 🚀 Feature

Selected params are not visible in chart popover when there is grouping applied.

### Motivation

It would be great to always have manually selected params visible as they might have higher priority from user perspective.

### Pitch

Move user-selected params to the top in chart popover

PR #1740, opened 2022-05-11
[feat] Show user-selected params before group config in chart popover

roubkar:
- Show user-selected params before group config in chart popover
- Tune chart popover box-shadow


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1739

VkoHov (approved):
lgtm!

PR #1731, opened 2022-05-07
[feat] Add Base and Figures Explorers

rubenaprikyan:
This PR includes BaseExplorer' core functionalities,  visualizations, simple usage.

## Core
- [x] Pipeline(Data flow container)
- [x] Cache mechanism

# Pipeline phases
- [x] Create Adapter phase
- [x] Create Fetch phase
- [x] Create Grouping phase

## Visualization
- [x] Virtualized box visualizer
- [x] BaseExplorer UI compositor
- [x] Simple Usage 
- [x] Add base features(range sliders, improve code health for dynamic explorer generations)
- [x] Code clean-up
- [x] FiguresExplorer based on BaseExplorer

PR #1755, opened 2022-05-17
[feat]BaseExplorer adapter phase

rubenaprikyan:
Adapter phase is designed to transform data coming from Query phase into hash table to manipulate it later easily.

This PR includes
- Process decoded data
   - Create an object for every visualizable object(Image, Line, Figure, Distribution etc.) and add it to top level hash table
   - Collect modifier values(contexts, params, etc.) from objects
- Adapter includes default data processor with different object depths support and possibility to pass custom processor


KaroMourad (approved):
lgtm

rubenaprikyan:
Resolves #1731

PR #1759, opened 2022-05-17
[fix] Fix Autocomplete input value updating frequency

arsengit:
- Fixed value updating frequency of Autocomplete input


VkoHov (approved):
lgtm!

Issue #1760, opened 2022-05-17
Make multiple zooming the default option


VkoHov:
## 🐛 Bug

The default option of zooming is `single` now. So every time it's necessary to click on the `zoom-in` icon to zoom the chart by default. 

### Expected behavior

Zoom multiple times by clicking on the `zoom-in` icon once by default.

PR #1761, opened 2022-05-17
[fix] Change the zooming default option to multiple

VkoHov:
1. Change the zooming default option to multiple
2. Move the multiple zooming option to the top of the control popover


VkoHov:
Resolves #1760

arsengit (approved):

KaroMourad (approved):

Issue #1737, opened 2022-05-10
Overview tab on single run page not loading


justin-barton:
## 🐛 Bug

The overview tab on the single run page is not loading and presenting the text: "Something went wrong." (see screenshot below)
![Screenshot 2022-05-10 at 16 54 46](https://user-images.githubusercontent.com/48782327/167670955-b827caed-94e9-482b-80af-9e8c86490076.png)

### To reproduce

Attempt to load the overview tab on any single run, for example:
http://xx.xx.xx.xx:43800/runs/d67ed8ad0d7346188e0cca6c/overview

### Expected behavior

Page should load.

### Environment

- Aim Version 3.9.2
- Python 3.8.10
- pip 22.0.4
- Ubuntu 20.04.3 LTS
- Chrome Version 101.0.4951.54 (Official Build) (x86_64)

### Additional context

Console errors:
```
react-dom.production.min.js:216 Error: Trying to redefine non-empty obj['yaml']
    at v._fill (index.js:118:15)
    at v._fill (index.js:114:10)
    at v._fill (index.js:114:10)
    at v._fill (index.js:114:10)
    at index.js:159:12
    at Array.forEach (<anonymous>)
    at v.object (index.js:153:20)
    at Function.object (index.js:90:31)
    at RunOverviewTabParamsCard.tsx:18:22
    at Object.useMemo (react-dom.production.min.js:172:115)
```

arsengit:
Hey @justin-barton, this fixing has been shipped with v3.10.0.
To upgrade please run pip install aim --upgrade

gorarakelyan:
closing the issue as it was resolved. @justin-barton please feel free to reopen the issue in case of any errors wrt run page loading.

PR #1767, opened 2022-05-18
[fix] Fix run overview tab params card error

arsengit:
- Fix run overview tab params card error in the single run page
- Fix Autocomplete setting previous state issue.


arsengit:
Resolves #1737

KaroMourad (approved):

Issue #1778, opened 2022-05-20
All the requests get stuck in pending state for long running requests when Live Update is enabled


rubenaprikyan:
## 🐛 Bug

When request execution time is more than 1-2s, all the requests get stuck in pending state.

### To reproduce

- Open Metrics Explorer
- Enable network throttling, or generate large amount of logs
- Enable live update

### Expected behavior

Live update should sent requests after each other.

### Environment

- Aim Version 3.10

### Additional context

Need to create a smart scheduler for live update requests

rubenaprikyan:
hey @ashutoshsaboo !
Tagging you because of #1654.  If you have a chance, please test this one, and let me know how it looks, then I'll close this issue.

PR #1785, opened 2022-05-21
[fix] Fix issue with live update requests scheduler

rubenaprikyan:
- Added the next requests scheduler after the last one will be handled (decoded, parsed)
- Increased live update delay to 10000
- Added a code block for backward compatibility on changing live update delay


rubenaprikyan:
Resolves #1778

VkoHov (approved):
lgtm!

Issue #1798, opened 2022-05-25
Remove logged system parameters from Explorer pages


alberttorosyan:
## 🚀 Feature

Currently `Run` parameters include both user-defined and auto-collected ones i all contexts. Need to keep system parameters only in the context of single Run page. 

### Motivation

Depending on the environment collected system parameters might be much more than user-defined, worsening the search/browsing UX and negatively affecting load performance.

### Pitch

Remove the `__system_params` key from all APIs by default (except single run params). For all APIs provide a qparam to change the default behavior if needed.

### Alternatives
Push the skip-list down to the container level. skip rocksdb keys range matching the skip-list.

PR #1799, opened 2022-05-25
[feat] Hide system parameters by default

alberttorosyan:
Added `skip_system` qparam to all APIs collecting the `Run` params: `Run.get(...)`
By default system parameters are skipped, except `/<run.hash>/info/` endpoint.


alberttorosyan:
resolves #1798

mihran113 (approved):
lgtm

PR #1825, opened 2022-06-01
[feat] Increase visibility/usability of the Show table diff button

arsengit:
- Moved the `Show table diff` button from manage columns to the top panel of the table.
- Disable the `Show table diff` button if the columns are already hidden.


KaroMourad (approved):
lgtm

VkoHov (approved):
lgtm!

PR #1827, opened 2022-06-01
[refactor] Cleanup pipeline phases codebase and add status changes callbacks

rubenaprikyan:
- Beautified Query phase, add possibility to pass status changes listener callback
- Promisified adapter, add statusCallback into adapter
- Add E2E pipeline with statusChanges


rubenaprikyan:
Resolves #1731

PR #1838, opened 2022-06-05
[feat] Add BaseComponents and possibility to dynamically create new explorer engine

rubenaprikyan:
This PR includes
-  BaseComponents (QueryForm, Grouping, Visualizer)
-  an utility function to create Explorer by passing just configuration, this function will create components, provided by user or using default components,
   It's possible to pass all components using in explorer including, QueryForm, Grouping,  Visualizer(width multiple visualizations)
- configuration includes default boxStyles injection possibility, and the engine will create a state from the default box config, with update and reset methods


rubenaprikyan:
Resolves #1731

Issue #1837, opened 2022-06-05
Keep color persistence state after page reload in grouping popover.


VkoHov:
## 🐛 Bug

Persisting coloring resets after page reload 

### To reproduce

1. Change color persisting
2. Reload the page

### Expected behavior

Have to keep the changes.

PR #1840, opened 2022-06-06
[fix] Keep color persistence state after page reload

VkoHov:
- Saving color persistence state in URL
- Tune chart tooltip overlays on the select-form dropdown
- Hide reverse mode from grouping popovers


VkoHov:
Resolves [#1837](https://github.com/aimhubio/aim/issues/1837)

PR #1845, opened 2022-06-07
[feat] Add the date and duration of the run in the run page header

VkoHov:
Add the date and duration of the run in the run page header


VkoHov:
Resolves [#1848](https://github.com/aimhubio/aim/issues/1848)

roubkar (approved):

PR #1852, opened 2022-06-08
[feat] Added caching mechanism for sequence queries

mihran113:


alberttorosyan (commented):
overall looks good.
a suggestion which might help to improve the overall code quality:
what about wrapping the cache in a class which can be accessed from any of the proxy object/view classes? if it's initialized caching is turned on and can be used.

mihran113:
> overall looks good. a suggestion which might help to improve the overall code quality: what about wrapping the cache in a class which can be accessed from any of the proxy object/view classes? if it's initialized caching is turned on and can be used.

that might help, but we'll still need to pass down the `run_hash` to proxy and eager classes, to retrieve the correct subview of the cache, so still might be a little bit of a mess overall :)

mihran113:
Resolves #1868

alberttorosyan:
> > overall looks good. a suggestion which might help to improve the overall code quality: what about wrapping the cache in a class which can be accessed from any of the proxy object/view classes? if it's initialized caching is turned on and can be used.
> 
> 
> 
> that might help, but we'll still need to pass down the `run_hash` to proxy and eager classes, to retrieve the correct subview of the cache, so still might be a little bit of a mess overall :)  

fair point. lets move forward then. approving this PR

alberttorosyan (approved):

Issue #1902, opened 2022-06-22
Highlight the error position when getting search syntax errors


arsengit:
## 🚀 Feature

Display the error cursor/position when getting search syntax errors

### Motivation

Provide users a better experience with query input. After improvement users can easily see and fix the query error. 

### Pitch

After getting an error highlight the error position and helpful error message.

PR #1930, opened 2022-06-29
[feat] Display error position(cursor) when getting syntax errors

arsengit:
- Provided better experience working with searching.
- Display and highlight syntax errors after bad requests.


arsengit:
Resolves #1902

Issue #1818, opened 2022-05-31
Web ui errors provide no information 


manuel-delverme:
## 🐛 Bug
![image](https://user-images.githubusercontent.com/1309922/171069665-ee8c69da-b930-420d-9639-1d7c2e9ff790.png)

I can't give any more information

roubkar:
Hey @manuel-delverme, sorry for the inconvenience.
Could you please share the logs from the terminal?

SGevorg:
I saw this yesterday too on the Tr11 logs: http://3.134.118.187/runs/d18c5921e07a4ae197e79e41/logs

PR #1933, opened 2022-06-30
[fix] Handle generic Exceptions and adjust HTTPException handling

alberttorosyan:
Added new error handler to cover all `Exception` sub-classes. Used as a fallback
Handle case when no 'detail' is available


alberttorosyan:
resolves #1818

Issue #1920, opened 2022-06-28
More readable interface to query runs by date


gorarakelyan:
## 🚀 Feature

Expose an easier and more readable interface to query runs by date.

### Motivation

Currently runs can be filtered by creation date, which is specified in unix timestamps e.g. `run.creation_time >= 1637438400`. Exposing a more readable interface for querying/filtering runs by date would make search much easer.

### Pitch and Alternatives

1. One option could be adding `hour`, `day`, `week`, `month`, `year` properties on date/time fields of runs to access them in more readable format. For example: `run.creation_time.day > 5 and run.creation_time.month == 2`

2. An alternative option is to enable search by passing strings in e.g. `YYYY-MM-DD hh:mm:ss` format.

3. We can go even further by enabling formatting datetime fields like [moment.js](https://momentjs.com/), examples:
```
run.creation_time.format('MMMM Do YYYY, h:mm:ss a'); // June 28th 2022, 1:06:00 pm
run.creation_time.format('dddd');                    // Tuesday
run.creation_time.format("MMM Do YY");               // Jun 28th 22
run.creation_time.format('YYYY [escaped] YYYY');     // 2022 escaped 2022
run.creation_time.format();                          // 2022-06-28T13:06:00+04:00
```

Additionally, a `runs` object could be added to search context to enable search by the latest runs. Example: `run.hash in runs[:-4]`

mihran113:
So the proposed solution is: 
Add 2 new fields `created_at` and `finalized_at` on `run` objects which will represent run creation and finalization times as python `datetime` objects and  will enable to query runs as following:

1. `run.created_at.day > 5 and run.created_at.month == 2`
2. `run.created_at.strftime("%d/%m/%Y") == "24/12/2021"`
3. `run.finalized_at > datetime(2022, 2, 2)`

Basically any operations that can be done in `python` with `datetime` objects, will be available in queries as well.

PR #1948, opened 2022-07-06
[feat] Add more user friendly querying for dates

mihran113:
Added `created_at` and `finalized_at` fields for `aim.Run` which are represented as `datetime.datetime` objects


mihran113:
Resolves #1920

alberttorosyan (approved):
lgtm

PR #1951, opened 2022-07-07
[fix] Auto-Detect inconsistencies

mahnerak:
`SequenceV2Data` has access to both metadata and series data trees that are not necessarily based on the same physical storage.
Therefore, we don't have a consistency guarantee between the two.
The implemented methods should tolerate this by taking what's older.

Addresses #1985


alberttorosyan (approved):
please update CHANGELOG.md

PR #1952, opened 2022-07-07
[fix] Prevent the active run indicator appearance out of the chart

VkoHov:
Prevent the active run indicator appearance out of the chart


KaroMourad (approved):
lgtm

PR #1960, opened 2022-07-12
[feat] Use style variable for font-weight across Aim UI

arsengit:
- Use style variable for font-weight across Aim UI
- Fix input width in manage columns popover
- Set the cell font-size to 15px for the tables in the RunDetail page


VkoHov (approved):
lgtm!

PR #1961, opened 2022-07-13
[feat] Add `duration` property for `Run` objects

mihran113:


alberttorosyan (commented):
looks good. should we also expose the new property in web APIs?

mihran113:
> looks good. should we also expose the new property in web APIs?

Not sure, no such requests have been made, and also the same logic is applied in the UI, this is explicitly for queries for now

mihran113:
Resolves #1964

alberttorosyan (approved):
lgtm

PR #1966, opened 2022-07-13
[feat] Update format of the query on searching by date

arsengit:
- Add `created_at`, `finalized_at`, and `duration` keys to the autocomplete suggestions list.
- Update format of the query on searching by date


aim/web/ui/src/services/api/api.ts
```diff
@@ -148,6 +145,17 @@ function remove<ResponseDataType>(url: string, options?: RequestInit) {
   });
 }
 
+function getTimezoneOffset(): string {
```
KaroMourad, line 148: do we really need this function, the body of function is too small and doesn’t reused it anywhere, IMO it would be better to remove it but if you want to reuse it later try to keep it in utils dir

roubkar (approved):

arsengit:
Resolves #1920

Issue #1997, opened 2022-07-25
Long metric name doesn't look good on UI


Sharathmk99:
## 🐛 Bug

We have a long metric name which is really long, but UI looks names are overlapping. Possible to handle somehow?

![image](https://user-images.githubusercontent.com/3970340/180880862-36daa2e4-96f2-4903-a7ad-f408611cc427.png)

### To reproduce

Log metrics with long name which is more than 256 char.
### Environment

- Aim Version (e.g., 3.0.1)
- Python version
- pip version
- OS (e.g., Linux)
- Any other relevant information

roubkar:
Hey @Sharathmk99, looking into this and will try to fix it ASAP.

Sharathmk99:
@roubkar Thank you.

gorarakelyan:
@Sharathmk99 the issue has been fixed and released with Aim v3.12.1 🚀

Sharathmk99:
Thank you so much @gorarakelyan!!

PR #2001, opened 2022-07-26
[fix] Break long metric names into multiple lines in Run Page

roubkar:
- Break long metric names into multiple lines in Run Page
- Fix a minor issue with the selection pane shadow for the Explorer table


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1997

arsengit (approved):

PR #2014, opened 2022-07-31
[feat] CLI Convert Tensorboard to accept remote server

### LINKED ISSUE #2013: aim converter cli command should accept remote server address while logging tensorboard event files
## 🚀 Feature

aim converter cli to log tensorboard logs should accept aim remote server repo address. Currently it only supports local path

### Motivation

aim cli converter should accept aim remote server address to log the artifact.

### Pitch

Support tensorboard convert cli command to accept remote server address

Sharathmk99:
Fixed #2013

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Enhancements:
 
 - Restrict Run.hash to auto-generated values only (alberttorosyan)
+- CLI option convert tensorboard to accept remote server address (sharathmk99)
 
 ### Fixes:
 
--- a/aim/cli/convert/commands.py
+++ b/aim/cli/convert/commands.py
@@ -2,27 +2,41 @@ import os
 
 import click
 from click import ClickException
-
+import grpc
 from aim.sdk.repo import Repo
-from aim.sdk.utils import clean_repo_path
+from aim.ext.transport.client import Client
 from aim.cli.convert.processors import (
     parse_tb_logs,
     parse_mlflow_logs
 )
 
+def validate_repo(ctx, param, value):
+    if not value and not param.required:
+        return value
+    if value.startswith("aim://"):
+        try:
+            _client = Client(value.replace("aim://", ""))
+            _client.get_version()
+            return value
+        except grpc.RpcError as e:
+            raise e
+    else:
+        validate_path = click.Path(exists=True,
+                file_okay=False,
+                dir_okay=True,
+                writable=True)
+        return validate_path.convert(value, param, ctx)
 
 @click.group()
-@click.option('--repo', required=False, type=click.Path(exists=True,
-                                                        file_okay=False,
-                                                        dir_okay=True,
-                                                        writable=True))
+@click.option('--repo', required=False, callback=validate_repo)
 @click.pass_context
 def convert(ctx, repo):
     ctx.ensure_object(dict)
-
-    repo_path = clean_repo_path(repo) or Repo.default_repo_path()
+    if not repo:
+        repo_path = Repo.default_repo_path()
+    else:
+        repo_path = repo
     repo_inst = Repo.from_path(repo_path)
-
     ctx.obj['repo_inst'] = repo_inst
 
 
--- a/aim/cli/convert/processors/tensorboard.py
+++ b/aim/cli/convert/processors/tensorboard.py
@@ -1,5 +1,6 @@
 import json
 import os
+from pathlib import Path
 
 import click
 from tqdm import tqdm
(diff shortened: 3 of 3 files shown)


alberttorosyan:
@Sharathmk99 thanks for the contribution 🙌 
Let me share a coupe of thoughts regarding the implementation.
One of the primary reasons remote server was not supported by converter at first place is the high frequency of I/O (due to `track()` method calls), which remote tracking server is not designed for. So you might experience big latency, which is not seen at first glance since the `track()` calls are non-blocking.
The way we think about the implementation is the following. Run the converter with local Aim repo. Implement a command (i.e. `aim rsync`) to copy the newly created runs to the remote Aim repo. This can be simplified even further by doing the conversion and sync as a single step. 
As of now, the `.aim` directory should be copied manually and `aim runs cp` command should be used. The alternative would be just copying the tb logs to the server where Aim remote repo resides and run the converter there.

Would love to hear your thoughts regarding this.

Sharathmk99:
> @Sharathmk99 thanks for the contribution 🙌 Let me share a coupe of thoughts regarding the implementation. One of the primary reasons remote server was not supported by converter at first place is the high frequency of I/O (due to `track()` method calls), which remote tracking server is not designed for. So you might experience big latency, which is not seen at first glance since the `track()` calls are non-blocking. The way we think about the implementation is the following. Run the converter with local Aim repo. Implement a command (i.e. `aim rsync`) to copy the newly created runs to the remote Aim repo. This can be simplified even further by doing the conversion and sync as a single step. As of now, the `.aim` directory should be copied manually and `aim runs cp` command should be used. The alternative would be just copying the tb logs to the server where Aim remote repo resides and run the converter there.
> 
> Would love to hear your thoughts regarding this.

I don’t think offline sync helps us. I’ll share my usecase, in big company normally tensorboard is used to log metrics. When they want to move to 3rd party tools first implementation is to sync existing tensorboard logs without much change real time and evaluate the tool. 
I think syncing tensorboard logs real-time is really good starting point for anyone who is already using tensorboard. 
First I started to make charges in convert cli, then I need to implement feature to real-time sync part of run itself. 
I think we need to seriously build backend API with high I/O support or no need to call backed API for every track call. When track is called, log the details temporarily to local and background thread to sync batch wise to server. We can talk about this, I have some ideas
If you think this is not the case for many other users, I’ll close the PR.

alberttorosyan:
Hi @Sharathmk99,

The use-case with tb converter you described above totally makes sense. The idea with syncing data with remove in batches seems to be the best option. Unfortunately implementation of batch sync with remote tracking server will require extensive changes in aim core and at this moment we do not have capacity to do that.
 
In order to make improve user experience we suggest the following:

1. First time sync of logs has high I/O rate and it worth considering to do it with local aim repo anyways.
2. Once the sync is done, aim repo can be moved to the destination server.
3. The rest of sync operations (for new training runs) should be done by one of the following two options:

**Option 1**
Monkey-patch tb logging to track data with aim remote tracking in addition to saving tb logs. With minimal changes in training scripts we can achieve lowest possible e2e latency for newly logged data

**Option 2**
Poll tb logs for new events (f.e. using cron). Remote tracking functionality should be extended to keep events cache on server side instead of local directory.

Would love to hear your thoughts regarding this.

Sharathmk99:
Hi @alberttorosyan 

Thank you for the response. It makes sense about the option you suggested.
In our team, we run many training per day and MLOps guys can't do above task manually. I understand there is latency consent when using remote server for syncing tensorboard logs.

But I think there is no harm in supporting remote server for tensorboard cli, so that end-user can use if it make sense for there use-case. For our use-case we didn't observe any latency problem syncing large number of metrics, as our K8S cluster is on-premise.

Let me know if you think there is huge problem supporting remote server for tensorboard sync. Many thanks!!

alberttorosyan:
> Hi @alberttorosyan
> 
> Thank you for the response. It makes sense about the option you suggested. In our team, we run many training per day and MLOps guys can't do above task manually. I understand there is latency consent when using remote server for syncing tensorboard logs.
> 
> But I think there is no harm in supporting remote server for tensorboard cli, so that end-user can use if it make sense for there use-case. For our use-case we didn't observe any latency problem syncing large number of metrics, as our K8S cluster is on-premise.
> 
> Let me know if you think there is huge problem supporting remote server for tensorboard sync. Many thanks!!

@Sharathmk99, implementing the support for remote server in tensorboard converter shouldn't be a big deal. We need to make the events file caching more abstract (currently it's bound directly to FS) so the Remote counterpart can be also implemented.

gorarakelyan:
Closing this PR. We are restructuring the remote+storage part and the conversion will be automatically supported for RT soon.

PR #2028, opened 2022-08-02
[feat] Support manylinux2014 wheels

alberttorosyan:


mihran113 (approved):

mihran113:
maybe worth changing the base branch to `release/3.12.x`? So the `3.12.1`  release will also have the manylinux2014 wheels.

Issue #2034, opened 2022-08-05
The metrics explorer doesn't seem to work


ilisparrow:
## 🐛 Bug

Hello, 
The metric explorer always says no results when I am sure it should, and It doesn't work for any metric. It doesn when I look at metrics in a run.

### To reproduce

Here are a few screenshots to show you what my problem is : 
![image](https://user-images.githubusercontent.com/4880273/183093110-93c0a7d0-dd10-480b-8aeb-c5c49b584e73.png)
When I go into a random run in the run explorer : 
![image](https://user-images.githubusercontent.com/4880273/183093320-be308fac-db8a-4a7b-a1a6-30713bc2e90f.png)

This is empty also, it used to be populated with the exact same code)
![image](https://user-images.githubusercontent.com/4880273/183101026-b5fc7816-bdcf-4d47-b4f0-c0f3f7626815.png)

### Expected behavior

Expected behavior is to get the plots displayed 

### Environment

- Aim Version 3.12.0
- Python 3.8.8
- pip 21.0.1
- Ubuntu 22.04

### Additional context

the CLI prints ; 
No index was detected

Thank you very much in advance

roubkar:
Hey @ilisparrow, thanks for reporting this, looking into it.
Meanwhile could you please share example logs to help us reproduce this 🙏

ilisparrow:
Thanks for the fast reply, 
Which logs do you me to share ? 
I track floats, integers and Text. 

I sometimes get this error  : 
```

┌------------------------------------------------------------------------┐
                Aim UI collects anonymous usage analytics.
                        Read how to opt-out here:
    https://aimstack.readthedocs.io/en/latest/community/telemetry.html
└------------------------------------------------------------------------┘
Running Aim UI on repo <Repo#2677591850585676717 path=/home/main/.aim read_only=None>
Open http://127.0.0.1:43800
Press Ctrl+C to exit
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/main/anaconda3/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/main/anaconda3/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/main/anaconda3/lib/python3.8/site-packages/fastapi/applications.py", line 199, in __call__
[...]

PR #2038, opened 2022-08-05
[fix] Correct metric column key in run table grid

VkoHov:
Format metric keys for run table grid columns correctly


KaroMourad (approved):

VkoHov:
Resolves [#2034](https://github.com/aimhubio/aim/issues/2034)

Issue #1479, opened 2022-03-08
in UI, seaborn heatmaps not displaying properly in figures


BasCR-hub:
## 🐛 Bug

Seaborn heatmap not displaying properly in figures. In the UI, the figure is blank in the 'Figures' tab for the run.

### To reproduce

```import matplotlib.pyplot as plt
import seaborn as sns
from aim import Run, Figure
f, ax = plt.subplots(figsize = (8, 8), dpi = 75)
sns.heatmap(np.random.rand(10, 10),
                   ax=ax)

run = Run()
aim_figure = Figure(f)
run.track(aim_figure, name="sns_heatmap", step=0) 
```

### Expected behavior

The heatmap should display properly, like it does for 'vanilla' plotly figures.

### Environment

OS: Ubuntu 20.04
Python : 3.8.5
aim==3.6.3
seaborn==0.11.1
matplotlib==3.4.2
plotly==5.1.0

devfox-se:
Hi @BasCR-hub thanks for reporting this!

We completely rely on Plotly to do the heavy lifting and unfortunately Plotly uses deprecated interface to try and perform the conversion, in fact conversion using latest versions for matplotlib, plotly and seaborn throw an exception. We're currently investigating possibility of direct integration of matplotlib. But first we have to verify there is an existing JS library available that can be used in UI.

I'll keep updating once we have a progress on this. Meanwhile you can use `aim.Image` object.

gorarakelyan:
@BasCR-hub @ashutoshsaboo it turned out there is an issue with converting seaborn and matplotlib figures to plotly, which is used to visaulize interactive plots on UI. The recommended way of tracking matplotlib figures is converting them into image via `aim.Image`.

gorarakelyan:
Closing this issue.

PR #2044, opened 2022-08-09
[feat] Warn to use Image if Figure fails

tmynn:


tmynn:
resolved #1479

aim/sdk/objects/figure.py
```diff
@@ -49,6 +53,8 @@ def _from_matplotlib_figure(self, obj):
             raise ModuleNotFoundError('Plotly is required to track matplotlib figure.')
 
         try:
+            logger.warning('Tracking a matplotlib object using "Figure" might not behave as expected.'
```
gorarakelyan, line 56: What about just restricting tracking matplotlib figures with aim.Figure? If the majority of figures are not supported by plotly converter, it might make sense. @tmynn

aim/sdk/objects/figure.py
```diff
@@ -49,6 +53,8 @@ def _from_matplotlib_figure(self, obj):
...
+                           'In such cases, consider tracking with "Image".')
```
gorarakelyan, line 57: worth putting here "aim.Image" to make it more obvious

gorarakelyan (approved):
lgtm

alberttorosyan (approved):

Issue #2043, opened 2022-08-09
Aim UI won't load in Colab


muellerzr:
## 🐛 Bug

Seems that aim's UI interface for Colab seems to be broken right now 😢 

### To reproduce

1. `!pip install aim`
2. `%load_ext aim`
3. `%aim up`

Error code of:
```bash
Launching Aim ...


Perhaps this is a bug from aim side.
Please open an issue https://github.com/aimhubio/aim/issues.
```

### Expected behavior
I'd expect to see the Jupyter interface running in Colab

### Environment

- Aim Version (e.g., 3.0.1) 3.12.2
- Python version 3.7.13
- pip version 21.1.3
- OS (e.g., Linux) Linux
- Any other relevant information

### Additional context

Tried running in both a CUDA and non-CUDA environment (though I fully know that likely means nothing)

gorarakelyan:
@muellerzr thanks for reporting! @rubenaprikyan could you please look into this?

rubenaprikyan:
@muellerzr thanks for reporting, looking into it, I'll get back here with updates.

rubenaprikyan:
Hi @muellerzr, the issue was fixed! 
Please try the latest version of  the package. And please lmk if you are still unhappy with `aim` by using in `colab`

muellerzr:
@rubenaprikyan now I get a 404 😓 Though this may not entirely be your fault, Colab just borked a ton of things over the weekend

rubenaprikyan:
@muellerzr have you tried to click on the link(Open Aim UI in a new browser tab)? If it works, the problem comes from the colab.

muellerzr:
@rubenaprikyan that also gives me a 404 it seems 😞

rubenaprikyan:
@muellerzr lets try one more thing, could you please add a / at the end of the url you opened in a new tab?

muellerzr:
@rubenaprikyan here's the URL it gave:

https://h79v8pd7no1-496ff2e9c6d22116-43801-colab.googleusercontent.com/notebook/

(So there already was a `/`).

Also tried adding a second / to no avail.

rubenaprikyan:
@muellerzr 
This is so weird. Yes, seems this is not related to the starting point of this issue, since the problem was already figured out and solved.

Thanks for sharing, I'll try to understand the source(colab or Aim) of this one, then will share the updates here.

rubenaprikyan:
Hey @muellerzr !
[...]

PR #2045, opened 2022-08-10
[fix] Fix issue on rendering aim ui on notebooks

rubenaprikyan:
This problems came after adding warnings in `FileSystemInspector`.
The problem is,  the manager of the aim server running on notebook extension listens to the `stderr` to detect aim server is living or not to show the ui, and warnings are corrupting `stderr`.
Added `--log-level=error` argument while running `aim up` command, and the aim cli will log only errors.
Warnings aren't mandatory for the notebook user, since they are not using terminal to run aim.

### DIFF
--- a/aim/cli/manager/manager.py
+++ b/aim/cli/manager/manager.py
@@ -16,7 +16,7 @@ class ManagerActionStatuses(enum.Enum):
 
 class ManagerActionResult:
     """
-        Object to returned by manager action
+        Object returned by manager action
         If status is ManagerActionStatuses.Failed the info dict should have a message property
         If status is ManagerActionStatuses.Succeed the info dict should have required properties for the specific action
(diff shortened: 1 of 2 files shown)


rubenaprikyan:
Resolves #2043

aim/cli/manager/manager.py
```diff
@@ -45,6 +45,7 @@ def run_up(args):
         'host': 'http://' + args['--host']
     }
 
+    time.sleep(2)
```
mihran113, line 48: why the sleep was moved here?

mihran113 (approved):

aim/cli/manager/manager.py
rubenaprikyan: It need to sleep a while before the server will be live, and no need to read `stderr` in that time

PR #2046, opened 2022-08-10
[feat] Run unit-tests before nightly releases

mihran113:
Added a job before packaging workflow to run unit tests for nightly releases. 
This is added as an extra job not a step, to be able to re-run without triggering the `version-override` job.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -28,6 +28,35 @@ jobs:
           git commit -m "Update VERSION file"
           git push
 
+  run-checks:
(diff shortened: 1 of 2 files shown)


alberttorosyan (commented):
looks good
what about including this step for minor and patch releases as well? 🤔

mihran113:
> looks good what about including this step for minor and patch releases as well? 🤔

Cause minor and patch releases are still done manually and this will increase time for the whole procedure, and we wanted to taste for nightlies first, if everything goes well, will add for other releases as well :)

alberttorosyan (approved):

Issue #1990, opened 2022-07-22
Ability to pin plots to the top in run detail page


gorarakelyan:
## 🚀 Feature

Enable a way to pin specific metric plots of a given run to the top in run detail page.

### Motivation

When checking a run results, there are some metrics that are looked at first. Instead of scrolling down to find those metrics each time, it would be great to pin them to the top.

### Pitch

Add pin icon on the plots. When clicking on the icon, the selected plot will be pinned to the top.

PR #2049, opened 2022-08-11
[feat] Adjust styles for metric card pin button

roubkar:
Moved pin button to top right corner of the metric card and added outlined styles to the button.


roubkar:
Resolves https://github.com/aimhubio/aim/issues/1990

arsengit (approved):

PR #468, opened 2021-05-12
Add Keras-Tuner integration

Khazhak:
Added Keras-Tuner integraation. I found a way to make a Callback as most of the other integrations, without creating a package as we planned earlier.

Closes #422.

PR #2052, opened 2022-08-11
[feat] Add Keras-Tuner integration

### LINKED ISSUE #422: Integration with Keras Tuner
Hi, I was wondering if this integrated with the tf Keras hyperparameter tuner and ,if so, are there any examples you could point me to? Thank you, great project.

tmynn:
Added Keras-Tuner integration for aim v3. This is an update of #468.


tmynn:
Resolves #422.

mihran113 (changes requested):
@tmynn  as we need the users to have an option to extend the callbacks, it's better to inherit the `AimCallback` directly from `TunerCallback` class. It's done this way with other adapters/callbacks.
You can have a look at already existing ones:
https://github.com/aimhubio/aim/blob/main/aim/sdk/adapters/pytorch_lightning.py

gorarakelyan:
@tmynn also pls do not forget to add docs

aim/sdk/adapters/keras_tuner.py
```diff
@@ -0,0 +1,66 @@
...
+    def on_batch_end(self, batch, logs=None):
+        if logs:
+            context = {'trial_id': self._current_trial_id}
+            for log_name, log_value in logs.items():
+                self._run.track(log_value, name=log_name, context=context)
+
+    def on_epoch_end(self, epoch, logs=None):
+        pass
+
+    def __del__(self):
```
alberttorosyan, line 64: I'm not sure if we need this at all. Does the callback provide a graceful `close()` method? 🤔
If yes, we can do it there. Also @mahnerak had some ideas of adding the resource cleanup to all supported adapters.

aim/sdk/adapters/keras_tuner.py
```diff
@@ -0,0 +1,66 @@
...
+        self._started_trials: List[str] = []
+        self.trial = None
+        self._run = None
+
+    @property
+    def experiment(self) -> Run:
+        if self._run is not None:
+            return self._run
+
+    def on_epoch_begin(self, epoch, logs=None):
```
gorarakelyan, line 36: What about creating a separate run for each trial and group them by an experiment? So:
- aim.Experiment - keras-tuner run
- aim.Run - a single trial, loging trial metrics and hparams

tmynn: @gorarakelyan that's exactly how we do.

aim/sdk/adapters/keras_tuner.py
tmynn: @alberttorosyan just removed the unnecessary method. @mahnerak do you have anything to add?

aim/sdk/adapters/keras_tuner.py
gorarakelyan: ah, I was confused seeing the trial_id context key on metrics. Then probably it is worth removing it from the context and attaching to run? run['trial_id'] = TRIAL_ID

the rest looks great!!

examples/keras_tuner_track.py
```diff
@@ -0,0 +1,64 @@
...
+    return tf.cast(record["image"], tf.float32) / 255.0, record["label"]
+
+
+train_ds = train_ds.map(standardize_record).cache().batch(64).shuffle(10000)
+test_ds = test_ds.map(standardize_record).cache().batch(64)
+
+tuner.search(
+    train_ds,
+    validation_data=test_ds,
+    callbacks=[AimCallback(tuner=tuner, repo='./aim_logs', experiment='keras_tuner_test')],
```
gorarakelyan, line 60: worth setting repo to cwd, no? `repo='.'`

aim/sdk/adapters/keras_tuner.py
tmynn: Sure, totally makes sense.

alberttorosyan (approved):

gorarakelyan (approved):
lgtm

mihran113 (approved):
lgtm

Issue #2057, opened 2022-08-12
Live update doesn't work after restarting aim server.


rubenaprikyan:
## 🐛 Bug

When live update is enabled, and aim server is down a while, it doesn't start working after restarting the server.

### To reproduce

- Enable live update
- Shut down the aim server for a while
- Then restart the server

### Expected behavior

The live update should start show results when aim server is down for a while, and restarted.

### Environment

- Aim Version (e.g., 3.0.1)  3.12.0
- Python version
- pip version
- OS (e.g., Linux)
- Any other relevant information

PR #2058, opened 2022-08-12
[fix] Add retry logic for failed network requests in live update

rubenaprikyan:
Added a retry for the failed network requests.


rubenaprikyan:
Resolves #2057

roubkar (approved):

PR #2069, opened 2022-08-16
[feat] Add the ability to sync sliders of stacked boxes

KaroMourad:
- Add `sync` property to the `useDepthMap` custom hook
- Fix resetting depth map values on box config change

### DIFF
--- a/aim/web/ui/src/hooks/useDepthMap.tsx
+++ b/aim/web/ui/src/hooks/useDepthMap.tsx
@@ -9,7 +9,7 @@ import _ from 'lodash-es';
  * @param deps {unknown[]} - the dependency array, after changing the item of the array will reset the depth map. The default value is: []
  * @param sync {boolean} - it enables the depth map synchronization mechanism. The default value is: false
  *
- * return [depthSelector, onDepthMapChange] {IUseDepthMap}
+ * return [depthSelector, onDepthMapChange] {UseDepthMap}
  */
 
 interface IUseDepthMapProps<T> {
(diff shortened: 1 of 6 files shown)


KaroMourad:
Resolves #1731

roubkar (approved):

aim/web/ui/src/hooks/useDepthMap.tsx
```diff
@@ -6,43 +6,59 @@ import _ from 'lodash-es';
...
 
 interface IUseDepthMapProps<T> {
   data: Array<T>;
   groupItemCb: (item: T) => string;
   state: any;
   deps: unknown[];
+  sync?: boolean;
 }
 
+type IUseDepthMap = [
```
rubenaprikyan, line 23: Please use `I` for only interfaces

aim/web/ui/src/hooks/useDepthMap.tsx
```diff
@@ -6,43 +6,59 @@ import _ from 'lodash-es';
...
+  const grouped = _.groupBy(data, groupItemCb);
+
+  const generateMapBy = React.useCallback((value: number = 0) => {
+    const newDepthMap: Record<string, number> = {};
+    for (let [groupId, items] of Object.entries(grouped)) {
+      const maxValue = items.length - 1;
+      newDepthMap[groupId] = value > maxValue ? maxValue : value;
+    }
+    return newDepthMap;
+  }, deps);
```
rubenaprikyan, line 44: Why dependencies comes from outside?, you know everything `data`, `groupedItemCallback` in your memoized function

rubenaprikyan (changes requested):

aim/web/ui/src/hooks/useDepthMap.tsx
KaroMourad: we don't need to update `generateMapBy` function every time when data changed as it only works when `deps` updates

rubenaprikyan (approved):
Lgtm!

PR #2074, opened 2022-08-17
[feat] Adjust default box configs

VkoHov:
- Set the min-height and min-width to 200
- Set default height and width to 400


KaroMourad (approved):
lgtm

roubkar (approved):

PR #2080, opened 2022-08-18
[fix] Fix figures explorer crashing issue

KaroMourad:
- Need to check `run` existence before using it
- Remove `Figures` from `appnameEnum` to avoid the compile time errors


rubenaprikyan (approved):
Lgtm!

PR #2081, opened 2022-08-18
[feat] Add ability to change group values order

roubkar:
- Fix issue with ordering group items
- Add toggle switcher for grouping order to grouping popover


rubenaprikyan (approved):
Lgtm!

PR #2089, opened 2022-08-19
[feat] Add grace period interval to Run status watcher

alberttorosyan:


mihran113 (approved):
the changes look good, but please check the unit tests

PR #2096, opened 2022-08-22
[feat] Add support for metrics query with last value

mihran113:


gorarakelyan:
@mihran113 docs?

mihran113:
> @mihran113 docs?

Added ✅

docs/source/using/search.md
```diff
@@ -180,6 +180,7 @@ When iterating over metrics, use the `metric` keyword which represents the track
 | -------- | ----------- |
 | `name` | Metric name |
 | `context` | Metric context dictionary |
+| `values.[OPTION]` | Metric aggregated data about metric values such as `last`, `last_step` and etc. |
```
gorarakelyan, line 183: should it be `metric.values.last_step` or `metric.steps.last`?

gorarakelyan (approved):

docs/source/using/search.md
gorarakelyan: and if the only supported option currently is `last`, then probably it is worth considering putting just `values.last`

alberttorosyan (approved):
looks good!

gorarakelyan (approved):
lgtm

Issue #2099, opened 2022-08-23
Always display run color in the table


VkoHov:
## 🐛 Bug

The run color display only if the chart is grouped by color

### Expected behavior

The color boxes should always be visible.

PR #2101, opened 2022-08-23
[fix] Adjust the visibility of the color of the run in the table

VkoHov:
- Fixed the color badge when there is no grouping applied
- Always showing the color badge


VkoHov:
Resolve [#2099](https://github.com/aimhubio/aim/issues/2099)

roubkar (approved):

Issue #2087, opened 2022-08-19
Ability to log a dictionary of values at once


muellerzr:
## 🚀 Feature

Be able to do:

```python
mets = {"accuracy":0.72, "f1":0.99}
run.track(mets, epoch=epoch, context={"subset":"validation"})
```

### Motivation

There are multiple cases where the context and epoch of logging something is the same, but I'd want to log multiple values at once. E.g. metrics for example. 

### Pitch

I'd like to be able to pass in a dictionary of values that can be logged under the same context all at once

### Alternatives

Iterating over the values in a for loop:

```python
mets = {"accuracy":0.72, "f1":0.99}
for k,v in mets.items():
    run.track(v, name=k, epoch=epoch, context={"subset":"validation"})
```

PR #2103, opened 2022-08-24
[feat] Support dictionaries as argument of `Run.track`

alberttorosyan:
- Changed the value validation logic and `_track` implementation.
- Added unit-tests to cover dict tracking + negative cases with specifying `name` argument.
- Changed `examples/pytorch_track.py` to use the new interface.

### DIFF
--- a/aim/sdk/tracker.py
+++ b/aim/sdk/tracker.py
@@ -85,7 +85,7 @@ class RunTracker:
     def __call__(
         self,
         value,
-        name: str,
+        name: str = None,
         step: int = None,
         epoch: int = None,
         *,
@@ -107,7 +107,7 @@ class RunTracker:
         self,
         value,
         track_time: float,
-        name: str,
+        name: str = None,
         step: int = None,
         epoch: int = None,
         *,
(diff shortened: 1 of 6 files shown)


alberttorosyan:
resolves #2087

aim/sdk/tracker.py
```diff
@@ -116,23 +116,18 @@ def _track(
...
             ctx = Context(context)
-            seq_info = self.sequence_infos[ctx.idx, name]
-            if not seq_info.initialized:
-                self._init_sequence_info(ctx.idx, name, val)
-            step = step if step is not None else seq_info.count
+            for name, val in values.items():
+                seq_info = self.sequence_infos[ctx.idx, name]
+                if not seq_info.initialized:
+                    self._init_sequence_info(ctx.idx, name, val)
+                step = step if step is not None else seq_info.count
```
mihran113, line 126: doesn't this logic need to be changed, if dictionary is provided?

alberttorosyan: nope. all the operations, including getting sequence info, updating, etc. are done on the level of dict item, which is the same as it was before.
so if you track a dict of 5 items, 5 sequence infos should be checked.

mihran113 (approved):
lgtm

Issue #2105, opened 2022-08-25
Handle deprecations in PyTorch Lightning 1.7 API


djwessel:
## Proposed refactoring or deprecation

Change imports in `aim.sdk.adaptors.pytorch_lightning` to handle deprecations in PyTorch Lightning API.

* `pytorch_lightning.loggers.base.rank_zero_experiment` -> `pytorch_lightning.loggers.logger.rank_zero_experiment`
* `pytorch_lightning.loggers.base.LightningLoggerBase` -> `pytorch_lightning.loggers.logger.Logger`

### Motivation

Some of the API in PyTorch Lightning regarding custom loggers has been deprecated in versions 1.7. The imports should be adjusted accordingly.

gorarakelyan:
@djwessel thanks for reporting. Will make sure to release the fix with the upcoming patch releases.
btw would be happy to help in case you want to contribute!

djwessel:
@gorarakelyan Hey, I opened a PR with the fix as it was rather quick to do. Let me know if there is anything that needs to be fixed.

gorarakelyan:
Hey @djwessel. The fix has been shipped with Aim 3.13.1 version! Congrats on the first contribution! 🚀

PR #2107, opened 2022-08-25
[fix] Deprecation issues with PL v1.7

djwessel:
Change imports in `aim.sdk.adaptors.pytorch_lightning` to handle deprecations in PL v1.7.

Fixes issue #2105

### DIFF
--- a/aim/sdk/adapters/pytorch_lightning.py
+++ b/aim/sdk/adapters/pytorch_lightning.py
@@ -3,10 +3,19 @@ from typing import Any, Dict, Optional, Union
 from argparse import Namespace
 
 try:
-    from pytorch_lightning.loggers.logger import (
-        Logger,
-        rank_zero_experiment,
-    )
+    import pytorch_lightning as pl
+
+    if pl.__version__ < "1.7":
+        from pytorch_lightning.loggers.base import (
+            LightningLoggerBase as Logger,
+            rank_zero_experiment,
+        )
+    else:
+        from pytorch_lightning.loggers.logger import (
+            Logger,
+            rank_zero_experiment,
+        )
+
     from pytorch_lightning.utilities import rank_zero_only
 except ImportError:
     raise RuntimeError(
--- a/aim/sdk/adapters/pytorch_lightning.py
+++ b/aim/sdk/adapters/pytorch_lightning.py
@@ -5,7 +5,10 @@ from argparse import Namespace
 try:
     import pytorch_lightning as pl
 
-    if pl.__version__ < "1.7":
+    def versiontuple(v):
+        return tuple(map(int, (v.split("."))))
+
+    if versiontuple(pl.__version__) < (1, 7):
         from pytorch_lightning.loggers.base import (
             LightningLoggerBase as Logger,
             rank_zero_experiment,
--- a/aim/sdk/adapters/pytorch_lightning.py
+++ b/aim/sdk/adapters/pytorch_lightning.py
@@ -3,8 +3,8 @@ from typing import Any, Dict, Optional, Union
 from argparse import Namespace
 
 try:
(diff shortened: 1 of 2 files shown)


alberttorosyan:
@djwessel thanks for your contribution 🙌.
Since the new logger interface is introduced in PL `v1.7`, this change will break integration with lower version of PL.
It would be great if we check the PL version installed, and select the base class of AimLogger dynamically.

djwessel:
@alberttorosyan Good point. I've added a version check to the imports and tested on PL v1.7 vs v1.6 and seems to be working as expected.

alberttorosyan (approved):
Looks great! One last comment, please change the section in CHANGELOG.md to `Fixes`.

aim/sdk/adapters/pytorch_lightning.py
```diff
@@ -3,10 +3,19 @@
 from argparse import Namespace
 
 try:
-    from pytorch_lightning.loggers.base import (
-        LightningLoggerBase,
-        rank_zero_experiment,
-    )
+    import pytorch_lightning as pl
+
+    if pl.__version__ < "1.7":
```
gorarakelyan, line 8: what about using packaging.version for comparison of versions?
```py 
from packaging import version
version.parse(pl.__version__) < version.parse("1.7.0")
```

gorarakelyan:
worth considering merging to `release/3.13.x` branch instead of `main` to ship the fix with the next patch release

aim/sdk/adapters/pytorch_lightning.py
djwessel: I was trying to avoid adding additional requirements. So I guess that decision is up to you guys. If you would prefer it done this way, I will update the pull request.

gorarakelyan: Yea, it seems packaging is not included in PSL. I just worry string comparison will introduce potential issues, e.g. `'1.17.1' < '1.7.1'` will return `True`. What about cloning `Version` class from packaging lib to Aim source code? Or there might be better options. Thoughts?

djwessel: Good point. There are also many other possible edge cases depending on how strict they are with their versioning scheme.  The best way to ensure compatibility would be to use the packaging library. See this [question](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python) for more options.

gorarakelyan: @djwessel sorry for a late reply. It seems converting to a tuple and then comparing tuples is a better way to go. Except for there might be some exceptions with comparison of alpha/release/dev version. I would go with this option probably.  thoughts?

```py
def versiontuple(v):
    return tuple(map(int, (v.split("."))))

>>> versiontuple("2.3.1") > versiontuple("10.1.1")
False
```

alberttorosyan:
@gorarakelyan @djwessel changing base branch back to `main` to avoid large conflicts. Will add to the `release/3.13.x` after merge.

alberttorosyan:
@djwessel, I've also pushed changes suggested by @gorarakelyan. thanks a lot for your time and contribution 🙌

gorarakelyan (approved):

aim/sdk/adapters/pytorch_lightning.py
```diff
@@ -3,10 +3,22 @@
...
 
 try:
-    from pytorch_lightning.loggers.base import (
-        LightningLoggerBase,
-        rank_zero_experiment,
-    )
+    import pytorch_lightning as pl
+
+    def versiontuple(v):
+        return tuple(map(int, (v.split("."))))
```
mahnerak, line 9: The version name is not guaranteed to be consisted only of dots and numbers.

Issue #2115, opened 2022-08-29
`aim reindex` command fails if Repo has stalled runs


alberttorosyan:
## 🐛 Bug

`aim reindex` command fails with the error

### To reproduce

- start a training process with aim logger.
- kill the process
- rim `aim reindex`

```
This command will try to reindex all runs in aim repo located at '/Users/alberttorosyan/aimhub/aim'. This process might take a while. Do you want to proceed? [y/N]: y
Finalizing stalled runs:   0%|                                                                                                                                                                       | 0/1 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "/Users/alberttorosyan/aimhub/aim/venv2/bin/aim", line 8, in <module>
    sys.exit(cli_entry_point())
  File "/Users/alberttorosyan/aimhub/aim/venv2/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/alberttorosyan/aimhub/aim/venv2/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/alberttorosyan/aimhub/aim/venv2/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/alberttorosyan/aimhub/aim/venv2/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/alberttorosyan/aimhub/aim/venv2/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/alberttorosyan/aimhub/aim/aim/cli/reindex/commands.py", line 38, in reindex
    index_mng.reindex()
  File "/Users/alberttorosyan/aimhub/aim/aim/sdk/index_manager.py", line 70, in reindex
    self._run(run_hash).finalize()
AttributeError: 'MaintenanceRun' object has no attribute 'finalize'
```

### Expected behavior

Successfully reindex stalled runs in repo.

alberttorosyan:
shipped with `3.13.1` release

PR #2116, opened 2022-08-29
[fix] Use correct method to set Run finalization time

alberttorosyan:


alberttorosyan:
resolves #2115

alberttorosyan:
resolves #2115

mihran113 (approved):

PR #2118, opened 2022-08-29
[deprecation] Warn about deprecation when custom Run.hash is set

alberttorosyan:


mihran113 (approved):
lgtm

YodaEmbedding:
Is there still a way load an existing run by hash in order to query or log to it?

alberttorosyan:
> Is there still a way load an existing run (by hash?) in order to query or log to it?

Absolutely. Existing runs are not affected, even when the deprecated interface is removed. The change is related to only newly created runs.

Issue #1967, opened 2022-07-14
Improve documentation about SSL support for `aim server` command 


mihran113:
## 📚 Documentation

There's a documentation lack about the SSL support for `aim server` command and the client side configuration for it to work properly.

PR #2122, opened 2022-08-29
[doc] Add section for remote tracking server SSL support

mihran113:


mihran113:
Resolves #1967

docs/source/using/remote_tracking.md
```diff
@@ -201,6 +201,16 @@ with torch.no_grad():
     print('Test Accuracy: {} %'.format(100 * correct / total))
 ```
 
+### SSL support 
+
+Aim Remote Tracking server can be set up to use SSL certificates to run on secure mode.
+To enable secure mode for server provide `--ssl-keyfile` and `--ssl-certfile` arguments to `aim server `command,
+where `--ssl-keyfile` is the path to the private key file for the certificate and `--ssl-certfile` is the path to the signed certificate.
```
gorarakelyan, line 208: Worth adding a link to CLI reference of the server command

docs/source/using/remote_tracking.md
```diff
@@ -201,6 +201,16 @@ with torch.no_grad():
...
+
+On client side set `__AIM_CLIENT_SSL_CERTIFICATES_FILE__` environment variable as the path to the PEM encoded root certificates for the secure channel establishment with the server.
```
gorarakelyan, line 210: Suggestion:

In order to establish a secure connection with the server, the client has to be configured accordingly.
Please set `__AIM_CLIENT_SSL_CERTIFICATES_FILE__` environment variable to the root directory, where PEM encoded certificates are located, e.g. 

```shell
export __AIM_CLIENT_SSL_CERTIFICATES_FILE__=/path/of/the/certs
```

gorarakelyan (approved):

PR #2124, opened 2022-08-30
[feat] Tune mlflow converter run properties

tmynn:
Changed the way of getting mlflow run id. Added the run_name and the description.


alberttorosyan (commented):
@tmynn, looks good. please change the base branch to `release/3.13.x` so we can include the fix in the upcoming patch release

alberttorosyan:
@tmynn you should also rebase the `feature/tune-mlflow-converted-run-props` onto release branch. It seems it was created on top of the `main`.

PR #2127, opened 2022-08-31
[doc] Add documentation for tracking multiple values at once

alberttorosyan:


alberttorosyan:
docs for #2087

docs/source/quick_start/supported_types.md
```diff
@@ -20,6 +20,32 @@ provided by the package:
 - [Text](#text-tracking-with-aim)
 - [Figure](#figure-tracking-with-aim)
 
+### Tracking multiple values
+Starting from `v3.14.0` Aim provides ability to track multiple values at once. If the context and epoch are the same,
```
gorarakelyan, line 24: Should the version be mentioned or are the docs already versioned by rtd?

docs/source/quick_start/supported_types.md
```diff
@@ -20,6 +20,32 @@ provided by the package:
...
+the dictionary can be passed and the keys will be treated as names for tracked values. For example the code snippet below
```
gorarakelyan, line 25: suggestion:

To track multiple values of different sequences but of the same context, just pass a dictionary of `<name: value>` pairs to the `track` method:

```py
aim_run.track({'accuracy': 98.2, 'loss': 0.001}, context={'subset': 'train'}, step=10, epoch=1)
```

Note: `name` argument should be set to `None` in this case.

further examples ...

docs/source/quick_start/supported_types.md
alberttorosyan: @gorarakelyan I think we should. It's a fast way to know when the feature first appears, regardless the current latest version.

gorarakelyan: Is it a common practice? 
Having properly versioned docs should be enough imo.

mihran113: @gorarakelyan for example official python docs do that:
https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat

mihran113 (approved):
lgtm

docs/source/quick_start/supported_types.md
alberttorosyan: @gorarakelyan @mihran113 if we decide to mention the version in the docs, we can choose the format, as for now such lines are in multiple places and there's no structured approach to this

gorarakelyan (approved):

Issue #1684, opened 2022-04-21
Remove params with no referring runs


devfox-se:
## 🚀 Feature

CLI command to cleanup repo from orphan parameters ( parameters with no referring runs )

### Motivation

This will reduce lookup load on underlying db since we remove unnecessary entries

PR #2130, opened 2022-08-31
[feat] Cleanup repo orphan params/sequences

mihran113:


aim/cli/storage/commands.py
```diff
@@ -113,3 +114,69 @@ def restore_runs(ctx, hashes):
...
+
+    # everything left in the `repo_paths` set is subject to be deleted
+    if not repo_paths:
+        click.echo('No orphan params were found')
+        return
+
+    # acquire index container to delete orphan paths
+    index_tree = repo._get_index_tree('meta', timeout=5).subtree('meta')
+
+    # start deleting with the deepest paths first to bypass the cases when parent path is deleted before the child
```
alberttorosyan, line 180: would it be too complex to optimize this part? i.e. if there are paths `('a', 'b', 'c'),  ('a', 'b', 'd'), ('a', 'b', 'e'), ...` and path `('a', 'b')`, delete just `('a', 'b')`

docs/source/refs/cli.md
```diff
@@ -200,6 +200,7 @@ __storage subcommands__
 | `upgrade 2to3`   | Upgrades legacy Aim repository from `2.x` to `3.0`.                                      |
 | `upgrade 3.11+`  | Update metric sequence data format for given runs. At least one run should be specified. |
 | `restore`        | Rollback `Run` to old metric format if run backup is available.                          |
+| `prune`          | Remove orphan params/sequences with no referring runs                                    |
```
alberttorosyan, line 203: worth describing what we mean by saying "orphan"

alberttorosyan: maybe even a small section in docs, describing a use-model (i.e. run moved/deleted, params overwritten, etc.)

alberttorosyan (commented):
overall looks good!
left a couple of comments

also lets test on a relatively big repo to see how long does the command take to run

aim/cli/storage/commands.py
mihran113: Can't think of a relatively simple approach now, also I don't think that it's worth the trouble, as we're doing this operation on a single `rocksdb` container, shouldn't take too long I guess

docs/source/refs/cli.md
mihran113: done ✅

alberttorosyan (approved):
lgtm

mihran113:
Resolves #1684

Issue #2132, opened 2022-09-01
Error when returning WriteInstructionsResponse with exception from remote server


jiyuanq:
## 🐛 Bug

```
Server is mounted on 0.0.0.0:9999
ERROR:grpc._server:Exception calling application: 0.1 has type float, but expected one of: bytes, unicode
Traceback (most recent call last):
File "/usr/local/lib/python3.8/dist-packages/aim/ext/transport/server.py", line 178, in run_write_instructions
self._verify_resource_handler(resource_handler, client_uri)
File "/usr/local/lib/python3.8/dist-packages/aim/ext/transport/server.py", line 209, in _verify_resource_handler
raise UnauthorizedRequestError()
aim.ext.transport.server.UnauthorizedRequestError
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
File "/usr/local/lib/python3.8/dist-packages/grpc/_server.py", line 443, in _call_behavior
response_or_iterator = behavior(argument, context)
File "/usr/local/lib/python3.8/dist-packages/aim/ext/transport/server.py", line 200, in run_write_instructions
return rpc_messages.WriteInstructionsResponse(
TypeError: 0.1 has type float, but expected one of: bytes, unicode
```

### To reproduce

It's related to https://github.com/aimhubio/aim/issues/2131
When I restart remote tracker during a training run, I noticed that the UnauthorizedRequestError is not correctly returned due to this bug.

It looks like a bug here: https://github.com/aimhubio/aim/blob/main/aim/ext/transport/server.py#L201 where "version" should actually be a string according to the protobuf definition.

What's confusing is that only in "ResponseHeader" (https://github.com/aimhubio/aim/blob/main/aim/ext/transport/proto/remote_tracking.proto#L80) the type of "version" is float, while in all other cases it's string, which might also need to be fixed

### Expected behavior

WriteInstructionsResponse's version should be set correctly

### Environment

Aim Version: 3.13.0
Python version: 3.8.10
OS (e.g., Linux): ubuntu 20.04

alberttorosyan:
@jiyuanq thanks for reporting this issue and detailed description 🙌 
@mihran113 the similar issue occurred in the past; can you please take a look?

mihran113:
Sure, will have look and get back with updates.

mihran113:
@jiyuanq thanks for the detailed report, you've done the heavy work here. 🙌 
I've made the changes, we'll ship them with the next patch release.

[...]

PR #2142, opened 2022-09-02
[fix] Resolve type mismatches in response headers

mihran113:


mihran113:
Resolves #2132

alberttorosyan (commented):
will this change introduce incompatibility between client and server of different versions?

alberttorosyan (approved):
lgtm

PR #2154, opened 2022-09-08
[feat] Display tags table on the home page

arsengit:
- Add tagsApi service
- Display tags table on the home page


aim/web/ui/src/pages/Home/components/HomeLeft/TagsCard/useTagsCard.tsx
```diff
@@ -0,0 +1,165 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  // memoized table data
+  const tableData = React.useMemo(() => {
+    if (tagsStore.data) {
+      return tagsStore.data.map(
+        ({ name, archived, run_count }: any, index: number) => {
+          return {
+            key: index,
```
KaroMourad, line 34: let's set key as combination of index and name

KaroMourad: or if the id of the tag is the name, it means the name is unique, we can use the name as key

KaroMourad (approved):
lgtm

Issue #2152, opened 2022-09-06
`Logs` tab on single run page doesn't load when heavy terminal logs are tracked 


mihran113:
## 🐛 Bug

When opening `logs` tab on single run page it doesn't show anything, furthermore when switching to another tab while it's loading the following error appears on terminal.
```
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
socket.send() raised exception.
```

### To reproduce
Run the following script, then open aim UI and navigate to the run's logs tab.
```
import random
heavy_log_line = ''.join([str(hash(random.random())) for _ in range(2000)])


from aim import Run

run = Run(capture_terminal_logs=True)

for i in range(2000):
    print(heavy_log_line)
```

### Expected behavior

Logs to visualize normally.

### Environment

- Aim Version  - 3.13.1
- Python version - 3.6.13
- pip version - 21.3.1
- OS - macOS 12.4

### Additional context

Reducing the initial log batch size might help as well.

PR #2161, opened 2022-09-11
[fix] Improve logs tabs rendering on long logs

VkoHov:
- Replace the List component with the Grid component of react-window


VkoHov:
Resolves [#2152](https://github.com/aimhubio/aim/issues/2152)

Issue #2135, opened 2022-09-01
Create a Dashboard page to provide a better onboarding experience


arsengit:
## 🚀 Feature

Create the `Dashboard` page.

### Motivation

Provide users with a better onboarding experience. 

### Pitch

Here are what should be displayed on the Home page
- [x] Existing experiments
- [x] Tags table
- [x] Release notes
- [x] Bookmarks list
- [x] Quick links to Aim docs
- [x] Provide quick links to query archived and active runs in the Runs explorer
- [x] In progress runs

PR #2165, opened 2022-09-13
[feat] Use new network service in Dashboard page

arsengit:
- Use new network service for fetching activity data on the Home page


arsengit:
Resolves #2135

roubkar (approved):

PR #2166, opened 2022-09-14
[feat] Add release notes api with new network service

arsengit:
Added release notes API using github's API to display on the `Dashboard Page`


arsengit:
Resolves #2135

roubkar (approved):

PR #2167, opened 2022-09-14
[fix] Fix the data live update handling in the Logs tab

VkoHov:
- Prevent calling of the live update before the other request is in pending state
- Change the overscan count of the virtualized list from 1000 to 100
- Change per request logs count to 200


roubkar (approved):

KaroMourad (approved):
lgtm

Issue #2164, opened 2022-09-13
Add the ability to handle exceptions for pipeline phases on Base explorers including Figures explorer


KaroMourad:
## 🚀 Feature

Add the ability to handle exceptions for all pipeline phases.

### Motivation

As there are no exception handlers for BaseExplorer pipeline phases, need to add the ability to catch errors and handle them, by creating corresponding exception instances on Base explorers including Figures explorer.
At first for QueryForm syntax error highlighting (query phase), and for other pipeline phases.

### Pitch

Need to create the `BaseError` class,  which is responsible for creating the same type of errors for Base explorers exceptions. 
Next need to create the `PipelineError` class, which is responsible for creating the same type of errors for pipeline exceptions. 
And create classes for all pipeline phases: `QueryError`, `AdapterError`, and `GroupingError`, which are responsible for the corresponding pipeline phases.
After creating the above-mentioned classes:
-  add exception handlers to every pipeline phase by using them
- handle pipeline execution errors on the BaseExplorer engine
- add the ability to set/get errors from the store

PR #2168, opened 2022-09-14
[feat] Add the ability to handle exceptions for pipeline phases on Base explorers

KaroMourad:
Add: 
- `BaseError` class, which is responsible for creating the same type of errors for Base explorers exceptions
- `PipelineError` class, which is responsible for creating the same type of errors for pipeline exceptions
- classes for all pipeline phases: `QueryError`, `AdapterError`, and `GroupingError`, which are responsible for the corresponding pipeline phases.

After creating the above-mentioned classes:
- add exception handlers to every pipeline phase by using them
- handle pipeline execution errors on the BaseExplorer engine
- add the ability to set/get errors from the store


KaroMourad:
Resolves #2164

Issue #2177, opened 2022-09-18
Replace the Spinner loader with a lighter one


VkoHov:
## 🚀 Feature

The current Spinner loader is working with JS too. It's heavy and freezes if the page is frozen. 

### Motivation

The new Spinner implemented using only the HTML and CSS will be lighter.

PR #2178, opened 2022-09-18
[feat] Replace the spinner loader with a lighter one

VkoHov:
The new Spinner is implemented using only the HTML and CSS which is lighter than the previous one.


VkoHov:
Resolves [#2177](https://github.com/aimhubio/aim/issues/2177)

KaroMourad (approved):
lgtm

arsengit (approved):

PR #2179, opened 2022-09-19
[fix] Disable run status report for remote tracking

alberttorosyan:
Run status reporting uses FS to log in progress runs heartbeat and read in `aim-watcher`
In case of remote tracking, the directory with remote Repo path is created. Added check to temporarily disable Run progress reporting in case of remote tracking.


alberttorosyan:
resolves #2186

Issue #2175, opened 2022-09-16
An explicit way to disable lines smoothing in the Metrics Explorer


gorarakelyan:
## 🚀 Feature

An explicit way to disable lines smoothing in the Metrics Explorer.

### Motivation

In order to disable smoothing over line charts in the Metrics Explorer, the coefficient should be set either to 1 or 0 (depending on the smoothing type). It would be better to add an explicit way of completely disabling the smoothing, as the current experience might be confusing.

### Pitch

Smoothing control to become an on/off toggle. Once enabled, lines will be smoothen by a default coefficient, e.g. 0.6 in case of exponential moving average (EMA). Otherwise, the smoothing will be disabled.

### Alternatives

Add a button in the config popover to disable the smoothing.

PR #2184, opened 2022-09-20
[feat] Add ability to disable smoothing explicitly

KaroMourad:
Smoothing control to become an on/off toggle. 
Once enabled, lines will be smoothened by a default coefficient, e.g. 0.6 in case of exponential moving average (EMA). Otherwise, the smoothing will be disabled.


KaroMourad:
Resolves #2175

arsengit (approved):

PR #2185, opened 2022-09-20
[feat] Add MXNet integration

tmynn:
Added MXNet Gluon logging handler


aim/sdk/adapters/mxnet.py
```diff
@@ -0,0 +1,156 @@
...
+                estimator.logger.info(msg.rstrip(', '))
+        self.batch_index += 1
+
+    @property
+    def experiment(self) -> Run:
+        if not self._run:
+            self.setup()
+        return self._run
+
+    def setup(self, estimator, args=None):
```
mihran113, line 128: `estimator` should be optional as well, as there's a call to `setup()` without any arguments

mihran113 (commented):
left a comment there, the rest looks good

aim/sdk/adapters/mxnet.py
tmynn: Fixed the missing types and made `estimator` optional. I think having a `setup()` without arguments in the `experiment` property is normal, because the `estimator` is needed to extract the parameters.

PR #2193, opened 2022-09-23
[feat] Add fallback to old locking mechanism via environment variable

mihran113:


alberttorosyan:
lets name variable `FORCE_USE_FILELOCK`
and atore the result of env. var. to not get it every time.

aim/storage/locking.py
```diff
@@ -141,6 +141,12 @@ def AutoFileLock(
         A timeout of 0 means, that there is exactly one attempt to acquire the
         file lock.
     """
+    from aim.sdk.configs import AIM_USE_FILELOCK_ONLY
+
+    if os.environ.get(AIM_USE_FILELOCK_ONLY):
```
mahnerak, line 146: We better move this after line 152. It will let us add a warning like a `Detected filesystem of type {fs_type}. Using filelock only because FORCE_USE_FILELOCK is set` which is good thing for debugging purposes.

Some users may wrongly think that they filesystem supports `UnixFileLock`s but eventually run into a problem. In that case, we'll see this warning message which will hint us about the problematic setup.

mihran113:
> lets name variable `FORCE_USE_FILELOCK` and atore the result of env. var. to not get it every time.

done ✅

alberttorosyan (approved):
lgtm

roubkar (approved):

mihran113:
Closing as obsolete. The new locking mechanism should handle it all.

PR #2196, opened 2022-09-25
[fix]  Fix line chart hovering issue occurring when metrics with "nan" values are tracked

KaroMourad:
Fixed line chart hovering issue occurs when metrics with "nan" values are tracked.

We don’t support rendering "nan" values in the line chart yet, we are filtering them in the data processing phase, and in some cases, the line values simply become empty, after filtering "nan" values, which is the reason for the bug.

Additional code style improvement:
 - refactor highlight and zoom enums sources


KaroMourad:
Resolves #2199

roubkar (approved):

VkoHov (approved):
lgtm!

Issue #2102, opened 2022-08-23
Persist explorer state through url on Figures and Base Explorers


rubenaprikyan:
## 🚀 Feature

Need to persist the explorer state on FiguresExplorer through URL and Local Storage

### Motivation

It's hard to create the same configuration every time after leaving the FiguresExplorer, and it would be great to send explorer configuration to someone via url. 
It would be great to make this feature extendible for all explorers made by BaseExplorer.

### Pitch

Need to make every state chunk configurable in BaseExplorer (controls, groupings etc.) to add a flag which will indicate the persistent storage type we want to use for the state chunk, and dynamically adjust values from storage(url, etc.) in initialization phase, and update storage(url, etc.) while updating the state.

### Additional context

There is similar functionality on all explorers.

Issue #2189, opened 2022-09-21
Figure explorer: URL-query string and bookmarks


YodaEmbedding:
## 🚀 Feature

Add URL-query string (e.g. `?figures=rate-distortion-plots&rowgroups=run.experiment&colgroups=run.hp.target_bitrate`) and bookmarks to Figure explorer.

### Motivation

Ability to refresh page and maintain same viewpoint, or share link to same Figure explorer layout with others.

Bookmarks may also be natural to apply here.

### Pitch

Some sort of URL query-string that points to the same UI view.

```
http://aimui:1234/figures?figures=rate-distortion-plots&rowgroups=run.experiment&colgroups=run.hp.target_bitrate
```

### Alternatives

N/A

### Additional context

N/A

rubenaprikyan:
Hey @YodaEmbedding! thanks for the request.
Our current milestone includes this enhancement and planned to ship with the next minor release.

rubenaprikyan:
Hey @YodaEmbedding !
The URL-query string feature has been shipped 🚀. Please check out the latest version.

The bookmark feature implementation is in progress.

PR #2214, opened 2022-09-29
[feat] Persist explorer state on Figures and Base Explorers

rubenaprikyan:
This PR includes changes on explorer core renderer and added new helpers, types, external package, and service.

**What is the main feature ?**
The main feature added with this pr is support to sync explorer state through url on core renderer and Figures Explorer
Added a `persist` flag on core renderer configuration for the whole explorer and `controls`, to identify does it need to persist or not?. Added initialization lifecycle method for all engines, to initialize them while initializing `explorer-engine`. Initializers are useful to register listeners or get default values from persistence storage(url | local storage) for the initial load. The engine initializer returns a function which works as de-initializer. Added a `name` and `basePath` arguments of `explorer-engine` creator function.
 
 **New package usage**
 Added a `history` package to imperatively handle  url.
 
**New Services**
Added a new service `browserHistory` over `history` package, which exports the history module and a new function `listenSearchParam` which is aimed to listen to a specific url search param updates. The usage of this service makes possible to keep engine's single responsibilities, i.e. `captionProperties` control is the owner of handling `captionProperties` search param. i.e. if control config has persist option provided as 'url', the initializer of the control's visualization will register listeners to listen url changes. also see the Pitch of #2102.

FYI also added a support to save box sizes configurations on device(local storage). The key generation process for local storage uses the name of the explorer (`Figures Explorer` => `figures-explorer`) as postfix. As the result the key should be [normalized name].visualizations.[visualization name].box. For the Figures Explorer the key is "figures-explorer.visualizations.vis1.box" and the value is encoded state.


rubenaprikyan:
Resolves #2102 #2189

roubkar (approved):

KaroMourad (approved):
lgtm

PR #2226, opened 2022-10-01
[fix] Fix displaying run count issue in the HeatMap cell

arsengit:
Fixed displaying run count issue in the HeatMap cell.


arsengit:
Resolves #2135

roubkar (approved):

PR #2227, opened 2022-10-02
[feat] Display nav link to tag page in case of empty data

arsengit:
Provide nav link to tag page in case of empty data


roubkar (approved):

arsengit:
Resolves #2135

PR #2228, opened 2022-10-02
[feat] Add Optuna integration

tmynn:
Added Optuna integration. A decorator for using Aim for tracking inside the objective function.


aim/sdk/adapters/optuna.py
```diff
@@ -0,0 +1,189 @@
...
+class AimCallback:
+    """Callback to track Optuna trials with Aim.
+
+    This callback enables tracking of Optuna study in Aim.
+    The study is tracked as a single experiment run,
+    where all suggested hyperparameters and optimized metrics
+    are logged and plotted as a function of optimizer steps.
+
+    .. note::
+        Users who want to run multiple Optuna studies within the same process
```
mihran113, line 25: I guess this case should be different with `aim`. Do I understand correctly that each study will be passed it's own `AimCallback` instance?

tmynn: No, the `AimCallback` instance is the same, we just create a new `aim.Run` for each trial.
Good point, we set the run_hash to None on line 132.  Maybe the docstring should be modified as:

"...should call ``aim.experiment.close()`` and set the aim.experiment._run_hash to None between subsequent calls to..."

What do you think?

mihran113: I don't think that users should access the private properties of the callback, maybe provide `close()` method on the callback itself, which will close the run, and reset the run hash, and mention that one in the docstring

aim/sdk/adapters/optuna.py
```diff
@@ -0,0 +1,192 @@
...
+                    system_tracking_interval=self._system_tracking_interval,
+                    log_system_params=self._log_system_params,
+                )
+                self._run_hash = self._run.hash
+
+        if args:
+            for key, value in args.items():
+                self._run.set(key, value, strict=False)
+
+    def close(self) -> None:
```
mihran113, line 164: let's add `self._run` existence check, as this can be called by the user even before `self._run` is initialized

mihran113 (approved):
left a small comment there, the rest looks good

aim/sdk/adapters/optuna.py
```diff
@@ -0,0 +1,192 @@
...
+
+        if not self._as_multirun:
+            self.setup()
+
+    def __call__(self, study: optuna.study.Study, trial: optuna.trial.FrozenTrial) -> None:
+
+        if isinstance(self._metric_name, str):
+            if len(trial.values) > 1:
+                # Broadcast default name for multi-objective optimization.
+                names = ["{}_{}".format(self._metric_name, i) for i in range(len(trial.values))]
```
alberttorosyan, line 87: lets use f-strings syntax

aim/sdk/adapters/optuna.py
```diff
@@ -0,0 +1,192 @@
...
+        experiment_name: Optional[str] = None,
+        system_tracking_interval: Optional[int] = DEFAULT_SYSTEM_TRACKING_INT,
+        log_system_params: bool = True,
+    ) -> None:
+
+        _imports.check()
+
+        if not isinstance(metric_name, Sequence):
+            raise TypeError(
+                "Expected metric_name to be string or sequence of strings, got {}.".format(
```
alberttorosyan, line 65: lets use f-strings syntax

aim/sdk/adapters/optuna.py
```diff
@@ -0,0 +1,192 @@
...
+
+            else:
+                names = [self._metric_name]
+
+        else:
+            if len(self._metric_name) != len(trial.values):
+                raise ValueError(
+                    "Running multi-objective optimization "
+                    "with {} objective values, but {} names specified. "
+                    "Match objective values and names, or use default broadcasting.".format(
```
alberttorosyan, line 97: same here

aim/sdk/adapters/optuna.py
```diff
@@ -0,0 +1,192 @@
...
+                        len(trial.values), len(self._metric_name)
+                    )
+                )
+
+            else:
+                names = [*self._metric_name]
```
alberttorosyan, line 103: should `names = self._metric_name` be enough?

aim/sdk/adapters/optuna.py
```diff
@@ -0,0 +1,192 @@
...
+        if self._as_multirun:
+            metrics["trial_number"] = trial.number
+
+        attributes = {"direction": [d.name for d in study.directions]}
+
+        step = trial.number if self._run else None
+
+        if not self._run:
+            self.setup()
+            self._run["run_name"] = f"trial-{trial.number}"
```
alberttorosyan, line 116: what about setting `Run.name` attribute?

aim/sdk/adapters/optuna.py
tmynn: As the `metric_name` is `Union[str, Sequence[str]]` we write a generic operation for all types of sequences.

aim/sdk/adapters/optuna.py
tmynn: Sounds good

aim/sdk/adapters/optuna.py
alberttorosyan: I mean just Line 96, not the whole logic of getting names.
as an Iterable `[*self._metric_name]` and `self._metric_name` are identical no?

alberttorosyan (approved):

aim/sdk/adapters/optuna.py
tmynn: Sure, sorry haven't notices the usage of the `names` variable on the further lines.

alberttorosyan (approved):
lgtm

Issue #2060, opened 2022-08-12
A like button for the run explorer


ilisparrow:
## 🚀 Feature

A like column to be able to highligh a specific model

### Motivation
When I run multiple experiments, Sometimes it's hard to know which run was selected as the one being deployed or the best without looking into notes.

### Pitch

Get a new column in the run explorer with a start or heart button that would be associated to a run

### Alternatives

be able to display notes in the run explorer

THank you very much in advance.

gorarakelyan:
@ilisparrow sounds great, what about displaying run tags in the runs explorer and enabling editing right from the table? Thus, an associated run may be tagged with `deployed` tag. Thoughts?

ilisparrow:
Hello, 
Thanks, This seems faire and would solve the problem. 
Best regards, 
Ilias.

VkoHov:
Hey @ilisparrow, this feature was released with v3.14.0, please try it out and let me know if it works as expected.

gorarakelyan:
@ilisparrow closing this issue as the enhancement has been shipped. Please feel free to reopen it in case of any issues. 🙌

Issue #2054, opened 2022-08-11
Add column for tags in Runs Explorer


djwessel:
## 🚀 Feature

Add a column in the Runs Explorer which shows which tags (if any) are associated with each run.

### Motivation

I want a very quick overview of which tags are associated with which runs while exploring the results of my experiments. You can go to the tags page to see the list of runs associated with a given tag, but this is out of the way and doesn't provide the overview information that you would get from including them in the Runs Explorer.

### Pitch

Add a new column called run.tags (as you can already filter by run.tags in the search bar) which visualizes all the associated tags.

VkoHov:
Hey @djwessel. The PR has been submitted - [#2120](https://github.com/aimhubio/aim/pull/2120). It is planned to be released in the 3.14.0 version. I will make sure to let you know once it is shipped.

VkoHov:
Hey @djwessel, this feature was released with v3.14.0, please try it out and let me know if it works as expected.

gorarakelyan:
@djwessel closing this issue as the enhancement has been shipped. Please feel free to reopen it in case of any issues. 🙌

PR #2236, opened 2022-10-03
[fix] Prevent unnecessary requesting the list of the tags

VkoHov:
- Remove the `)` symbol from the attached tags popover
- Prevent unnecessary requesting the list of the tags


roubkar (approved):

VkoHov:
Resolves https://github.com/aimhubio/aim/issues/2060, https://github.com/aimhubio/aim/issues/2054

PR #2237, opened 2022-10-03
[feat] Add guides docs links on the Dashboard page

arsengit:


aim/web/ui/src/config/references/index.ts
```diff
@@ -70,42 +72,82 @@ const GUIDES = {
   },
 };
 
+/*
+ getDocsVersion() returns the version of the docs to be used in the links
+ */
+function getDocsVersion() {
+  let splitVersion = `${AIM_VERSION}`.split('.');
```
KaroMourad, line 79: let's make this row more secure as there can be `rc`, `dev` versions in which cases this is incorrect to pop and push at the end of the string,
`let [majorVersion, minorVersion] = `${AIM_VERSION}`.split('.');`

KaroMourad (commented):
lgtm

arsengit:
Resolves #2135

PR #2238, opened 2022-10-03
[feat] Add outlined state on project statistics cards

KaroMourad:
Add outlined state on cards:
- when hovering over on the self card 
- when hovering over on a statistics bar item

Divide statistics cards highlighting states into two states:
- filled:  when the card is also a link (to the corresponding explorer)
- default: when the card has no link to navigate (there is no explorer yet)


KaroMourad:
Resolves #2135

roubkar (approved):

arsengit (approved):

PR #2248, opened 2022-10-06
[feat] Redesign the selection popover of the tags

VkoHov:
- Combine the select tag and attached tags popup
- Sort the tags list by the selected once
- Implement a search bar in tags popover


roubkar (approved):

KaroMourad (approved):
lgtm

PR #2298, opened 2022-10-27
[feat] Implement the blobs URI system engine in the base explorer

VkoHov:
- Implement the blobs URI system engine
- Add the blobs request creator method

### DIFF
--- a/aim/web/ui/src/modules/core/engine/blob-uri-system/index.ts
+++ b/aim/web/ui/src/modules/core/engine/blob-uri-system/index.ts
@@ -1,0 +1,163 @@
+import _ from 'lodash-es';
+
+import { createBlobsRequest } from 'modules/core/api/runsApi';
+
+import { throttle } from 'components/Table/utils';
+
+import { SequenceTypesEnum } from 'types/core/enums';
+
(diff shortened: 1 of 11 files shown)


aim/web/ui/src/modules/core/engine/blob-uri-system/index.ts
```diff
@@ -0,0 +1,159 @@
...
+    getBlobsData: (uris: string[]) => any;
+  };
+}
+
+const BATCH_SEND_DELAY = 1000;
+
+function createBlobURISystemEngine(
+  sequenceType: SequenceTypesEnum,
+): IBlobURISystemEngine {
+  let blobsData: Record<string, string> = {};
```
rubenaprikyan, line 30: Why those bindings are using let ?

rubenaprikyan (commented):
Lgtm!

aim/web/ui/src/modules/core/engine/blob-uri-system/index.ts
VkoHov: Changed the blobsData and blobsSubscriptions to const.

rubenaprikyan (approved):

roubkar (approved):

Issue #2297, opened 2022-10-26
Solid stroke mismatch in UI


twni2016:
## 🐛 Bug

The solid line shown in the graph, is displayed incorrectly as a dashed line in the table. 

### To reproduce

When I groupby the curves by stroke, the solid curve shown in the graph (see the screenshot below)

![Screen Shot 2022-10-26 at 5 06 14 PM](https://user-images.githubusercontent.com/20842215/198138734-dc54fa13-cb36-4374-81ef-109c168982c2.png)

is depicted as a dashed curve (the top column) in the table (see the screenshot below). Note that the top column's stroke is `style="stroke-dasharray: 12, 3, 6, 3, 3, 3, 3, 3;"` (which correspond to solid curve in the graph), and the bottom column's stroke is `style="stroke-dasharray: 12, 3, 6, 3;"` (which correspond to dashed curve in the graph).

![Screen Shot 2022-10-26 at 5 06 30 PM](https://user-images.githubusercontent.com/20842215/198138488-cdf59875-d225-4bf1-96bd-fdd932e3bd31.png)

### Expected behavior

The solid curve is shown as solid in the table. 

### Environment

- Aim Version 3.14.1
- Python version 3.10
- pip version 22.2.2
- macOS chrome

gorarakelyan:
@twni2016 thanks for reporting, this is proper weird. 
@roubkar is it even supposed to have `style="stroke-dasharray: 12, 3, 6, 3, 3, 3, 3, 3;"` stroke style?

roubkar:
We have a fixed list of `stroke-dasharray`s:

```javascript
[
  'none',
  '5 5',
  '10 5 5 5',
  '10 5 5 5 5 5',
  '10 5 5 5 5 5 5 5',
  '20 5 10 5',
  '20 5 10 5 10 5',
  '20 5 10 5 10 5 5 5',
  '20 5 10 5 5 5 5 5',
]
```

@twni2016 seems you're using a custom list for dasharrays.

Anyway, it looks like the problem is in `LineChart` as the solid line should be rendered as dashed.
Will look into this and keep you updated on the progress.

KaroMourad:
Since we are processing stroke styles before rendering stroke badges in the table, these` stroke-dasharray` 
`12, 3, 6, 3, 3, 3, 3, 3;`  value is valid. 

Have found an issue with the processing part. Already fixed the bug related to 'stroke-dasharray: none' values. 
[...]

PR #2299, opened 2022-10-27
[fix] Fix solid stroke rendering issue

KaroMourad:
Before rendering `stroke-dasharray` badges in the table, we are processing (decreasing) the stroke styles, to have the ability to draw badges in a short area.
 
 Fix:
 - pass the 'none' values of the `stroke-dasharrays` on processing stroke styles


KaroMourad:
Resolves #2297

arsengit (approved):

PR #2306, opened 2022-10-29
[fix] Fix querying all data on empty query

KaroMourad:
when navigating to `/metrics` all the metrics are retrieved instead of asking to select metrics


rubenaprikyan (approved):
Lgtm!

PR #2310, opened 2022-10-31
[feat] Add Distribution .from_histogram and .from_samples

### LINKED ISSUE #2183: Pre-binned distribution/histogram
## 🚀 Feature

A user should be able to create an `aim.Distribution` using histogram data that the user computed. Perhaps by specifying some flag that disables the automatic internal `numpy.histogram`.

### Motivation

Sometimes the exact histogram is known, rather than needing to a sample it from some random source.

### Pitch

```python
counts = [4, 2, 1]
bin_edges = [0, 9, 18, 99]  # or bin_midpoints?
aim.Distribution(counts=counts, bin_edges=bin_edges)
```

### Alternatives

Plotting a typical plot.ly figure instead.

### Additional context

N/A

YodaEmbedding:
Fixes #2183


alberttorosyan:
@YodaEmbedding thanks for opening PR! 🙌
Will look into changes shortly.

alberttorosyan (commented):
@YodaEmbedding the changes look good!
The only possible issue is that the changes in `Distribution.__init__` are not backwards compatible. This might break existing training scripts. Lets extend the `__init__` signature to cover both cases but adjust the documentation to advertise the class methods.

@rubenaprikyan @roubkar could you please check are there any changes required on the UI side?

YodaEmbedding:
The primary intent is to be able to specify either data `samples` or the `hist` frequency counts directly.

```python
# Either of these should display the same histogram:
samples = [0, 0, 0, 1, 2, 2, 2, 2, 3]
hist    = [3, 1, 4, 1]
```

I notice that, currently, only uniform bins are able to be represented. Thus, if a user specifies `bin_edges`, it will mostly be ignored anyways. With this in mind, perhaps I should eliminate `bin_edges` for now:

```python
def __init__(self, samples=None, bin_count=64, *, hist=None):
    if samples is None and hist is not None:
        if len(hist) != bin_count:
            raise ValueError("Length of hist must be same as bin_count.")
    ...
```

mihran113:
> The primary intent is to be able to specify either data `samples` or the `hist` frequency counts directly.
> 
> ```python
> # Either of these should display the same histogram:
> samples = [0, 0, 0, 1, 2, 2, 2, 2, 3]
> hist    = [3, 1, 4, 1]
> ```
> 
> I notice that, currently, only uniform bins are able to be represented. Thus, if a user specifies `bin_edges`, its interior contents are ignored anyways. With this in mind, perhaps I should eliminate `bin_edges` from the proposed signature for now:
> 
> ```python
> def __init__(
>     self, samples=None, bin_count=64,  # Original.
>     *,
>     hist=None  # Additionally proposed by this PR.
> ):
>     if samples is None and hist is not None:
>         if len(hist) != bin_count:
>             raise ValueError("Length of hist must be same as bin_count.")
>     ...
> ```

Totally makes sense 🙌 Let's go for this option now, then keep the more sophisticated solution for later on, as it requires changes on UI side as well.

mihran113 (approved):
Looks good to me. Let's see if @alberttorosyan has any additional comments.

alberttorosyan (approved):
looks good!

mihran113:
Hey @YodaEmbedding! As no blockers were left, updated the changelog and merged the changes. Thanks a lot again for contribution 🙌

Issue #2312, opened 2022-10-31
The group property name isn't displayed in the images and audio tabs


VkoHov:
## 🐛 Bug

The group property name isn't displayed in the images and audio tabs on the run page.

### To reproduce

1. Go to the Run page.
2. Switch to the Images tab.(Be sure that there are already tracked images)
<img width="1070" alt="Screen Shot 2022-10-31 at 20 37 42" src="https://user-images.githubusercontent.com/43929589/199061074-f37009ed-792f-479f-a089-715228ef2c86.png">

### Environment

- Aim Version 3.14.3

PR #2313, opened 2022-10-31
[fix] Fix the group property name visibility in the images and audio tabs

VkoHov:
Fix the group property name visibility in the images and audio tabs on the run page


VkoHov:
Resolves [#2312](https://github.com/aimhubio/aim/issues/2312)

arsengit (approved):

PR #2317, opened 2022-11-03
[fix] Use `packaging.version` to parse version strings

### LINKED ISSUE #2316: Pytorch Lightning Adapter expects versions to consist of `int`s
The [Pytorch Lightning Adapter](https://github.com/aimhubio/aim/blob/15e394cc39b9def1a283f7e1ca7e6aac7a6d2027/aim/sdk/adapters/pytorch_lightning.py) in 3.14.3 extracts a version tuple from the version string. It expects all parts to map to `int`:

```python
    def versiontuple(v):
        return tuple(map(int, (v.split("."))))
```

I noticed the problem when I received the following exception: `ValueError: invalid literal for int() with base 10: 'post1'`

I suggest using [`packaging.version.parse`](https://packaging.pypa.io/en/latest/version.html#packaging.version.parse), so I'll submit a PR soonish.

jangop:
Closes #2316.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Add epoch tracking for PyTorch Lightning (tmynn)
 - Add PaddlePaddle integration (tmynn)
 - Add Optuna integration (tmynn)
+- Use `packaging` to parse version strings(jangop)
 
 ## 3.14.3 Oct 29, 2022
 
--- a/aim/sdk/adapters/pytorch_lightning.py
+++ b/aim/sdk/adapters/pytorch_lightning.py
@@ -2,13 +2,12 @@ import os
 from typing import Any, Dict, Optional, Union
 from argparse import Namespace
 
+import packaging.version
+
 try:
     import pytorch_lightning as pl
 
-    def versiontuple(v):
-        return tuple(map(int, (v.split("."))))
-
-    if versiontuple(pl.__version__) < (1, 7):
+    if packaging.version.parse(pl.__version__) < packaging.version.parse("1.7"):
         from pytorch_lightning.loggers.base import (
             LightningLoggerBase as Logger,
             rank_zero_experiment,
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,7 @@ REQUIRED = [
     # fastapi to support python3.6
     'async-exit-stack>=1.0.0; python_version<"3.7"',
     'async-generator>=1.0; python_version<"3.7"',
+    'packaging>=15.0',
 ]
 
 if platform.machine() != 'arm64':


jangop:
@gorarakelyan that is entirely understandable. However, `packaging` is not just any dependency. You are using `setuptools` to package aim, and so, you already have access to `packaging`. From what I understand.

Personally, I consider `packaging` almost-stdlib and version string parsing is not something I think anyone should reimplement:  [PEP440](https://peps.python.org/pep-0440/).

alberttorosyan:
> @gorarakelyan that is entirely understandable. However, packaging is not just any dependency. You are using setuptools to package aim, and so, you already have access to packaging. From what I understand.

As I recall, `setuptools` is the build-time only dependency and with the isolated builds it should not affect environment where aim is installed.

> Personally, I consider packaging almost-stdlib and version string parsing is not something I think anyone should reimplement: [PEP440](https://peps.python.org/pep-0440/).

@jangop, fair point 🙌  @gorarakelyan, just checked the `packaging` package size, the universal wheel is only ~40KB, so maybe worth considering it 🤔 ?

CC @mahnerak

mahnerak:
We had this: https://github.com/aimhubio/aim/pull/2107#discussion_r973485036
Not sure why we didn't resolve it.

> setuptools is the build-time only dependency and with the isolated builds it should not affect environment where aim is installed.

That's right. 
`setuptools` and `packaging` are not stdlib, and are available for aim only during (isolated) build. The `pip` itself is able to compare versions, but even `pip` is not always available. 

> and version string parsing is not something I think anyone should reimplement: 

Totally agree. I think we better add a new dependency rather implementing our parser. Most of the python environments already have `packaging`, so even that `40kb` is rarely going to be installed just for `aim`.

jangop:
> Most of the python environments already have `packaging`, so even that `40kb` is rarely going to be installed just for `aim`.

That's a better version of what I was trying to say. 😅

jangop:
Oh, in the commit I did not add `setuptools` to `setup.py`, because I really do not know which version to pick.

gorarakelyan:
Makes a lot of sense. Let's just make sure to put really loose version boundaries to avoid dependencies incompatibilities (if there may be any) during installation.

jangop:
> Makes a lot of sense. Let's just make sure to put really loose version boundaries to avoid dependencies incompatibilities (if there may be any) during installation.

Do you / does anyone know of a tool to figure out the earliest version / point in time when a certain function / functionality was added to a given dependency? This is a problem I always struggle with.

alberttorosyan:
> > Makes a lot of sense. Let's just make sure to put really loose version boundaries to avoid dependencies incompatibilities (if there may be any) during installation.
> 
> Do you / does anyone know of a tool to figure out the earliest version / point in time when a certain function / functionality was added to a given dependency? This is a problem I always struggle with.

Not aware of any such tool... Usually I check the release notes, but in this case there's nothing mentioned... Though I've looked at the github repo; version `15.0` already has the `parse` function.

jangop:
> > > Makes a lot of sense. Let's just make sure to put really loose version boundaries to avoid dependencies incompatibilities (if there may be any) during installation.
> > 
> > 
> > Do you / does anyone know of a tool to figure out the earliest version / point in time when a certain function / functionality was added to a given dependency? This is a problem I always struggle with.
> 
> Not aware of any such tool... Usually I check the release notes, but in this case there's nothing mentioned... Though I've looked at the github repo; version `15.0` already has the `parse` function.

Thanks a million. I'll add it.

alberttorosyan (approved):

alberttorosyan:
@jangop, approved and merged PR. Thanks for contribution 🙌

Issue #2041, opened 2022-08-09
Add "Experiment page"


gorarakelyan:
## 🚀 Feature

Add an "Experiment page" to view overall experiment info, as well as explore and manage (e.g. delete, archive, etc) attached runs.

### Motivation

As the project gets bigger it becomes essential to keep runs organized. "Aim experiment" helps to group related runs together, but it lacks functionality to help to organize and manage groups of runs for large and long-lasting projects.

### Pitch

Add a separate page (like "Run page"), where would be displayed:
- experiment overview: runs count, crashed runs count, experiment start date, etc
- attached runs list

Enable quick actions for the experiment runs - "delete", "archive", etc.

Furthermore, enable leaving notes for an experiment.

PR #2330, opened 2022-11-07
[feat] Setup the experiment page layout and implement the header

VkoHov:
- Setup the experiment page layout
- Implement the navigation in the header 
- implement the tabs and content wrapper


VkoHov:
Resolves [#2041](https://github.com/aimhubio/aim/issues/2041)

aim/web/ui/src/pages/Experiment/useExperimentData.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { IExperimentData } from 'modules/core/api/experimentsApi';
+
+import experimentEngine from './ExperimentStore';
+
+function useExperimentData(experimentId: string) {
+  const { current: engine } = React.useRef(experimentEngine);
+  const experimentStore: IResourceState<IExperimentData> =
+    engine.experimentState((state) => state);
+  const experimentsStore: IResourceState<IExperimentData[]> =
+    engine.experimentsState((state) => state);
```
rubenaprikyan, line 13: `useExperimentsState` is the best name here thought

rubenaprikyan: The same for the other stores

aim/web/ui/src/pages/Experiment/useExperimentData.tsx
```diff
@@ -0,0 +1,35 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  React.useEffect(() => {
+    engine.fetchExperimentData(experimentId as any);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [experimentId]);
+
+  return {
+    experimentStore,
```
rubenaprikyan, line 29: What is the benefit of changing the name of `experiment(s)State` to `experiment(s)Store`?
As for the names `State` and `Stores`, I believe the engine should export the `Store` and the hook should export `State`. 
`State` - since the state, by the context of reactive programming is considered Observable by default
`Store` - since the store isn't observable in nature, it has some methods to get/subscribe to the Observable state and at the same time store has a lot of other functions, you can not render something directly from the store, I mean if we want to call some variable `State`, it should be considered as data, but the store isn't data at all.

i.e. I believe you don't use name Store as expected

rubenaprikyan (changes requested):

KaroMourad (approved):
lgtm

Issue #2347, opened 2022-11-15
Bring all history w/o sampling by cli convert wandb


hjoonjang:
## Proposed refactoring or deprecation

(Not sure if it's okay with this category: Feature suggestion instead..?)
In `aim/cli/convert/processors/wandb.py`, https://github.com/aimhubio/aim/blob/8441a58ac1e39b8cd5a5b802794c9c944effdfd4/aim/cli/convert/processors/wandb.py#L45
we've been trying bringing **SAMPLED** history of existing wandb metric logs.
I would carefully propose to bring **ALL** history without sampling.

It seems gonna be easily implemented using `wandb_run.scan_history()` with a little mod rather than using `wandb_run.history()`.
The latter has default args `samples=500` to have roughly 500 records returned. ([ref](https://github.com/wandb/wandb/blob/e85ba97972308091d22f34e275cd58fea4b774dc/wandb/apis/public.py#L2027))

### Motivation

In #2067, an awesome feature had been added: wandb to aim log converter.
But I felt lossless log transfers would be much more preferrable than the implicit sampling.
And it seems pretty easy to implement by replacing a single line of code.

### Pitch

Though [`wandb_run.history()`](https://docs.wandb.ai/ref/python/public-api/run#history) gives *simpler and faster* results, but it's sampled to loss some information eventually.
I propose it to be lossless fetch, which is easy to implement as well.

### Additional context

I've already had remote branch with fix commits, but after checking contributing guidelines, I've got to come here first.

---

By the way, I have a remaining question:
- I'm gonna have to add `import pandas` in `aim/cli/convert/processors/wandb.py`, but there is no `pandas` requirements in this project yet.
- I found it anyhow requires `pandas` package when I simply run the scripts below:
    ```shell
    pip install aim wandb
    aim convert wandb --entity ... --project ...
    # Error: "Unable to load pandas, ..." (from wandb api, not from aim, though)
    pip install pandas
    # Now it's working
    ```
    - By `wandb` itself: https://github.com/wandb/wandb/blob/e85ba97972308091d22f34e275cd58fea4b774dc/wandb/apis/public.py#L2058-L2063
        - Seems `wandb` also has no requirements asking `pandas`
- So, should `pandas` be gracefully added in requirements of this project?

gorarakelyan:
[...]

PR #2348, opened 2022-11-15
Fix cli convert wandb to bring all history w/o sampling

hjoonjang:
I've explained the details in #2347, which is resolved by this PR.

It makes cli convert wandb to bring **ALL** history, rather than implicitly **SAMPLED** history (roughly 500 by default).

---

My remaining question is: (also elaborated in #2347)

Should `pandas` be added on requirements?
I'm not sure if this adding up might harm any philosophy of this project, since I'm a newbie here 🤧..


gorarakelyan:
@hjoonjang thanks for the contribution! 
Probably a naive question, but is pandas dependency necessary here? 
It would be great to avoid having additional dependencies, as it may result in incompatibility issues between Aim dependencies and virtual env when installing Aim. As well as, make the installation process longer.

hjoonjang:
Thanks for your warm greetings!

I've made it avoid `pandas` dependency.
According to my local test, some skipping conditions have been added.

tmynn:
@hjoonjang thanks a lot for the proposed changes!

A question regarding the extra elif. Does the `test` appear in the key in your case? If yes was it added by some framework?
Also I think now we can avoid creating the "`records`" variable and directly iterate over the `run.scan_history()`.

The rest looks good to me.

hjoonjang:
@tmynn, thank you for suggesting ideas!

> @hjoonjang thanks a lot for the proposed changes!
> 
> A question regarding the extra elif. Does the `test` appear in the key in your case? If yes was it added by some framework? Also I think now we can avoid creating the "`records`" variable and directly iterate over the `run.scan_history()`.
> 
> The rest looks good to me.

I've applied what you asked: remove "`records`" variable and directly iterate.

Plus, concerning if it's expanding PR's scope, console log collector has been added on.
I've referred [this](https://github.com/aimhubio/aim/blob/c891aa038537b3b1ff019ca3d856f3822b8d22f1/aim/ext/resource/tracker.py#L222) part for "Logs" type tracking.

tmynn:
@hjoonjang thanks for your comment.
As for the  subsets, yes you can apply filtering based on subset. Please see: https://aimstack.readthedocs.io/en/latest/ui/pages/explorers.html#metrics-explorer
The script is initially created by having the average user experience in mind. Most of the users and frameworks use these key names.
The assumption that the key might contain a "`/`" character comes from fairseq/metaseq, which has a specific structure.

Of course one can have highly arbitrary key names. I believe the appropriate local customizations need to be made in such cases.
Please let me know if there is a general trend or pattern that you think need to be considered.

hjoonjang:
@tmynn, thanks for your kind indications.
I've missed the advanced search usage in the documentation. Now I see the straightforward utility of `subset` filtering. I'm (quite slowly) being updated with the philosophy of this project, thanks to your help.

So far, no special trend nor pattern has been come up with from my experience.
Please let me know if more changes have to be made!

tmynn:
@hjoonjang thanks for your contribution.
I agree with the TODOs, the initial script was designed to provide simple conversions, we will include media file conversion and system logs tracking as well in the future.
Please let us know if you would like to work on any of these and need our help.

tmynn:
@hjoonjang could you please pull from the `main` branch? The tests have just been updated.

hjoonjang:
@tmynn thanks for your help!
I've rebased to `main` this branch.

I'm interested in media file conversion from wandb logs.
I'll add up a new issue thread regarding this when I can work on it.

tmynn (approved):
lgtm

tmynn:
@hjoonjang looks good to me, merging it :blush:. Thanks for the contribution, would love to work on other features with you :rocket:

PR #2349, opened 2022-11-15
[feat] Robust Run locking mechanism

alberttorosyan:
### Background
Initial mechanism of locking was hinged on the Container abstraction (representing a single rocksdb instance). This causes some problems due to handling of multiple locks per Run.
The other layer of complexity is added by the fact that filelock type can be different depending on the underlying FS. Due to introduction of SoftFileLock in Aim, the Runs frequently stay in locked state, requiring manual intervention to "unlock", resume and optimize Runs.

The code change aims to solve problems above by:
1. Separation of Run locking, indexing and run active status.
2. Unify Run locks into singular SoftFileLock.
3. Provide robust mechanism to detect stalled Runs and indexing them without touching locks.
4. Provide mechanism (SDK and CLI) to forcefully release Run locks.
5. Provide mechanism to handle legacy locks and make transition to the new system smooth.

Remaining items:

- [ ] Improve `aim runs ls` command to list locks information (Run is open, last progress time, etc.)
- [x] e2e testing with complex scenarios.


alberttorosyan:
- Python script used for testing:
```python
import random
import time
import datetime
import os
import signal
import threading


from aim.sdk import Run
from aim.__version__ import __version__ as ver


def log(msg):
    now = datetime.datetime.now().time().isoformat(timespec='seconds')
    th_name = threading.current_thread().name
    print(f'[{now}] - {th_name} >>> {msg}')


run = Run()
log(f'starting Run \'{run.hash}\' with Aim \'{ver}\'.')
run.experiment = ver

PID = os.getpid()


def _kill():
    # sleep between 30 and 120 seconds
    sleep_interval = random.randint(30, 120)
    log(f'sleeping for {sleep_interval} seconds...')
    time.sleep(sleep_interval)

    # select random signal
    signals = (signal.SIGKILL, signal.SIGTERM, signal.SIGINT)
    sig = random.choice(signals)

    log(f'killing Run \'{run.hash}\' with signal {sig.name}.')
    os.kill(PID, sig)


killer = threading.Thread(name='KillThread', target=_kill, daemon=True)
killer.start()

log(f'tracking metrics for Run \'{run.hash}\'.')
for _ in range(1000):
    run.track(random.random(), name='m')
    run.report_progress(expect_next_in=60)
    time.sleep(1)
```

- Start script above in background:
`for run in {1..32}; do python _test_track_random.py & ; done`

- Start Aim UI

mihran113 (approved):
Works like a charm. One small comment there, the rest looks good.

alberttorosyan:
Resolves #2386

PR #2350, opened 2022-11-16
[feat] Add multi-flag support for check-ins

mahnerak:
Previously we had flag names for the `check-in`s but the they were not designed to separate those events.
The goal of this change is to redesign the writing process to make sure that independent throttling of multiple flags is possible. Independent and global flushing were also introduced.


alberttorosyan (commented):
looks good! please change the logging level for debug messages before merging to main.
need to update the `CHANGLELOG.md` and link to GH issue (open one if needed)

mahnerak:
Solves #2383

alberttorosyan (approved):
lgtm

PR #2364, opened 2022-11-18
[feat] Switch to patched version of official pynvml

mihran113:


alberttorosyan (approved):
@mihran113, let's change the version back and merge it to main.

Issue #2307, opened 2022-10-30
Implement the Audio explorer based on the base explorer. 


VkoHov:
## 🚀 Feature

Add the ability to search, filter, and compare audios of the runs using a common search mechanism.

### Motivation

The easy way to compare audios of the several runs.

PR #2372, opened 2022-11-22
[feat] Add Figures and Audios as options to Compare dropdown

KaroMourad:


KaroMourad:
Resolves #2307

aim/web/ui/src/modules/core/engine/pipeline/index.ts
```diff
@@ -131,7 +131,7 @@ function createPipelineEngine<TStore, TObject>(
       const queryState = store.getState().query;
 
       if (!queryState.ranges.isInitial) {
-        const url = updateUrlSearchParam(
+        const url = getUpdatedUrl(
           'query',
           encode({
             ...queryState,
```
rubenaprikyan, line 137: why this need to spread ?

rubenaprikyan (changes requested):

rubenaprikyan (approved):
Lgtm!

aim/web/ui/src/modules/core/engine/pipeline/index.ts
KaroMourad: there is only changed the name of the function

Issue #2374, opened 2022-11-22
UI hangon searching for runs


Natlem:
## 🐛 Bug

Unable to display runs or any kind of information from Aim running on a remote server.
The /runs page is stuck on "Searching over runs..."  with "34 of 34 checked ------ 33 matched run(s)" 
There's an error in the console log: 

> encoder.ts:25 Uncaught (in promise) DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range

server-side aim ui errors:
> query failed, name 'metric' is not defined
> query failed, '>' not supported between instances of 'datetime.datetime' and 'str'   
> query failed, 'NoneType' object is not callable  

### Environment

- Aim Version: 3.14.3 
- OS: Linux

mihran113:
Hey @Natlem! Could you please share what was the query that failed?

Natlem:
I think this should be all. Thanks for the great support :)

roubkar:
Hey @Natlem, the PR which resolves the issue related to characters encoding in Aim UI has been merged and will be released with the next version.

PR #2376, opened 2022-11-23
[fix] Support non-Latin chars for encoding in Aim UI

roubkar:
Use `encodeURI` and `decodeURI` methods along with `btoa` and `atob` to escape non-Latin characters


KaroMourad (approved):
LGTM

roubkar:
Resolves https://github.com/aimhubio/aim/issues/2374

PR #2380, opened 2022-11-23
[fix] Leave color information in tracked terminal logs

mihran113:
Fix csi(cursor state) handling: do not remove color information as it is handled by UI side.
 Remove from line only cursor up/down/right/left commands.

### DIFF
--- a/examples/pytorch_lightning_track.py
+++ b/examples/pytorch_lightning_track.py
@@ -103,7 +103,7 @@ def cli_main():
     # ------------
     # testing
     # ------------
-    trainer.test(test_dataloaders=test_loader)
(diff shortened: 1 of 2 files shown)


examples/pytorch_lightning_track.py
```diff
@@ -103,7 +103,7 @@ def cli_main():
     # ------------
     # testing
     # ------------
-    trainer.test(test_dataloaders=test_loader)
```
alberttorosyan, line 106: is this related to the fix?

mihran113: nope, just found out during the testing, and fixed it :)

alberttorosyan (approved):
lgtm

PR #2390, opened 2022-11-25
[fix] Add finalizers to all RT server objects

alberttorosyan:


aim/ext/transport/handlers.py
```diff
@@ -62,7 +89,8 @@ def get_lock(args: bytes):
         repo = Repo.default_repo()
     kwargs = decode_tree(unpack_args(args))
     run_hash = kwargs['run_hash']
-    return repo.request_run_lock(run_hash)
+    from aim.sdk.lock_manager import SFRunLock
```
mahnerak, line 92: Is importing SFRunLock necessary here?
If yes, please add comment to state it explicitly.

mahnerak (approved):
Added a reminder.
LGTM

Issue #2394, opened 2022-11-29
Incorrect empty-illustration message in Audios explorer


KaroMourad:
## 🐛 Bug

Audios explorer shows an incorrect message, when there is no tracked data, by mentioning the Figures explorer.

### To reproduce

1. do not track any data
2. open Audios explorer
3. check the empty-illustration message

### Expected behavior

Show a correct message for every explorer page, separately.

### Environment

- Aim Version (e.g., 3.15.0)

PR #2395, opened 2022-11-29
[fix] Fix showing incorrect empty-illustration message in Audios explorer

KaroMourad:
- create a dynamic `Illustration` component
- create a generic `getStaticContent` function for Base explorers, which will get the static content type and will return the corresponding empty-illustration or predefined static content
- create `getStaticContent` function for Base Explorers separately for Audios and Figures and set the function in the corresponding explorer configuration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## 3.15.1
 
 - Fix issue with index container lock for older repos (mihran113)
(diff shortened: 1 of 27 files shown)


KaroMourad:
Resolves #2394

rubenaprikyan (approved):
Lgtm

PR #2407, opened 2022-12-06
comment about ports and set `--workers  1`

flowpoint:
Running the dockerfile with the previous option `--workers 2` instead of 1 threw grpc errors in the python clients `Run(repo="...")` about the unreachable ports 43801 and 43802. I assume that those are additionally needed with port 43800 when running more workers. Publishing 43801 43802 seemingly fixed it with multiple workers when running the container.

I hope the note saves someone time


flowpoint:
Oh, this is rather related to 'aim server' not 'aim up' so the port issue needs to be addressed differently

mihran113:
Hey @flowpoint! Thanks for the notice, I guess this was due to the fact, that the scaling part wasn't yet included in the documentation, I've just merged those changes today.
https://aimstack.readthedocs.io/en/latest/using/remote_tracking.html#scaling
Now there's a section in the docs, that covers it. Hope it helps. 🙌

PR #2422, opened 2022-12-09
[fix] Track gpu stats even if gathering one stat fails

### LINKED ISSUE #2420: GPU utilization is not tracked if querying power usage fails
## 🐛 Bug

I am running experiments on a machine with a GPU, yet no GPU metrics are tracked. It seems like this code is responsible:

https://github.com/aimhubio/aim/blob/480e063cde063897283bcd8adb221e9baa861637/aim/ext/resource/stat.py#L152-L186

When any part of the GPU stats collection fails, we just give up entirely and store no information. In my case querying the power usage seems not supported by nvml, it raises `nvml.NVMLError_NotSupported`. Querying utilization and memory usage works just fine though and it would be nice if we could track those stats anyway.

### To reproduce

I'm not sure how to reproduce this, since it depends on a setup where `nvml` fails to determine the GPU power usage.

### Expected behavior

Aim tracks all the information that it can query without exceptions.

### Environment

- Aim Version (e.g., 3.15.1)
- Python version 3.10.9
- pip version 22.0.3
- OS (e.g., Linux) Linux
- Any other relevant information

### Additional context

--

timokau:
Fixes #2420. See the commit messages for additional details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Add other x-axis alignment and system logs tracking to cli convert wandb (hjoonjang)
 - Add support for pre-binned distribution/histogram (YodaEmbedding)
 
+### Fixes
+
+- Fix gpu stats logging when some stats are unavailable (timokau)
+
 ## 3.15.1 Dec 1, 2022
 
 - Fix issue with index container lock for older repos (mihran113)
--- a/aim/ext/resource/stat.py
+++ b/aim/ext/resource/stat.py
@@ -182,7 +182,7 @@ class Stat(object):
                     # Device temperature
                     'gpu_temp': round10e5(temp),
                 })
-        except Exception:
+        except nvml.NVMLError_NotSupported::
             pass
 
         return system, gpus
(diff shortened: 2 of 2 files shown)


timokau:
I get this error when I attempt to install the dependencies for the test suite:

```
$ pip install -r requirements.txt # in aim/tests
ERROR: ./aim/web/ui is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
```

alberttorosyan:
resolves #2420

timokau:
Done, I have update the changelog.

mihran113 (approved):
Looks good 🚀

alberttorosyan:
@timokau merged PR to `main`.
Thanks again for your contribution 🚀

PR #2444, opened 2022-12-22
[feat] Add group type property while creating hash key for the grouping

rubenaprikyan:
Added a type property of a hashed object as group key.


KaroMourad (commented):
lgtm

PR #2452, opened 2022-12-28
[feat] Add Stable-Baselines3 integration

tmynn:
Created a AimCallback class for initializing a Run, and an extra class AimOutputFormat for tracking the metrics. This solution is forced by sb3 design.

### DIFF
--- a/aim/sdk/adapters/sb3.py
+++ b/aim/sdk/adapters/sb3.py
@@ -81,9 +81,7 @@ class AimCallback(BaseCallback):
         self._run_hash = None
 
     def _init_callback(self) -> None:
-        args = {}
-        if "algo" not in args:
-            args["algo"] = type(self.model).__name__
+        args = {"algo": type(self.model).__name__}
         for key in self.model.__dict__:
             if type(self.model.__dict__[key]) in [float, int, str]:
                 args[key] = self.model.__dict__[key]
--- a/aim/sdk/adapters/sb3.py
+++ b/aim/sdk/adapters/sb3.py
@@ -1,0 +1,136 @@
+import logging
+import numpy as np
+from typing import Optional, Dict, Any, Union, Tuple
+
+from stable_baselines3.common.logger import KVWriter, Logger
+from stable_baselines3.common.callbacks import BaseCallback  # type: ignore
+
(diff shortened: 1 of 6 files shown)


aim/sdk/adapters/sb3.py
```diff
@@ -0,0 +1,136 @@
...
+    ):
+        self.aim_callback = aim_callback
+
+    def write(
+        self,
+        key_values: Dict[str, Any],
+        key_excluded: Dict[str, Union[str, Tuple[str, ...]]],
+        step: int = 0,
+    ) -> None:
+        for (key, value), (_, excluded) in zip(
```
alberttorosyan, line 32: lets check the existence/value of the `key` in `key_excluded` dict. it will allow to handle cases when `key_excluded` contains only the sub-set of actually excluded keys.

alberttorosyan (changes requested):

aim/sdk/adapters/sb3.py
tmynn: @alberttorosyan great idea.
Though just double-checked, those two dicts always have the same length and contain the same keys.
Here is an example of the keys and values of two dicts
- `key_values = {'train/learning_rate': 0.0007, 'train/n_updates': 99,...} `
- `key_excluded = {'train/learning_rate': None, 'train/n_updates': tensorboard,...} `

The keys that have a None value will be tracked.

aim/sdk/adapters/sb3.py
```diff
@@ -0,0 +1,136 @@
...
+        self.repo = repo
+        self.experiment_name = experiment_name
+        self.system_tracking_interval = system_tracking_interval
+        self.log_system_params = log_system_params
+        self._run = None
+        self._run_hash = None
+
+    def _init_callback(self) -> None:
+        args = {}
+        if "algo" not in args:
```
mihran113, line 85: @tmynn isn't this statement redundant?

tmynn: @mihran113 exactly! thanks for noticing

mihran113 (approved):

alberttorosyan (approved):
lgtm

PR #2458, opened 2023-01-04
Fix bug #2457

### LINKED ISSUE #2457: SDK `query_runs` fails with `TypeError`
## 🐛 Bug

While trying to run `repo.query_runs(query)`, followed by turning iterator into actual values, method fails with following traceback:

```
Traceback (most recent call last):
  File "temp.py", line 12, in <module>
    list(runs)
  File "/home/inc0/aim/aim/sdk/sequence_collection.py", line 235, in iter
    yield from run_seq
  File "/home/inc0/aim/aim/sdk/sequence_collection.py", line 122, in iter
    match = self.query.check(**{'run': run_view, seq_var: seq_view})
TypeError: check() keywords must be strings
```

### To reproduce
```
repo = Repo(repo_url)
runs = repo.query_runs(f"run.creation_time > {start_timestamp}")
list(runs)
```

### Expected behavior

Not throwing error

### Environment

- Aim Version - 3.15.2
- Python version - 3.8.15

### Additional context

After a bit of debugging I think issue is that [QueryRunSequenceCollection](https://github.com/aimhubio/aim/blob/main/aim/sdk/sequence_collection.py#L194) doesn't have `sequence_name` method. After I added this method, next error was about missing `allowed_dtypes`. This one is a bit harder as return type of this sequence is complex (`Run` objects, I think). I'm not sure what's best dtype to fill there.

inc0:
This commit adds missing methods to QueryRunSequenceCollection.

/Fixes #2457

### DIFF
--- a/aim/sdk/sequence_collection.py
+++ b/aim/sdk/sequence_collection.py
@@ -1,6 +1,6 @@
 import logging
 from abc import abstractmethod
-from typing import Iterator
+from typing import Iterator, Union, Tuple
 from typing import TYPE_CHECKING
 from tqdm import tqdm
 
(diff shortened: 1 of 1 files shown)


inc0:
seems like this actually doesn't fix it, just makes problem harder to spot

Issue #2207, opened 2022-09-27
Add grpc metadata to stub header


jennifer12121:
## 🚀 Feature
 Is it possible to add an extension to the current RemoteTrackingServiceStub class to allow for grpc information to be passed as metadata to the stub header? 

### Motivation
We are interested in being able to perform an access token call and with the current implementation, it does not allow metadata to be pass through the client.

### Additional Context

The RemoteTrackingServiceStub is used inside the client.py file: 

https://github.com/aimhubio/aim/blob/a87f7a7f5d867da41f686af3a51ce6639f75eeb0/aim/ext/transport/client.py#L55

alberttorosyan:
Hi @jennifer12121! Thanks for the request 🙌 
Would love to learn more details about your usage of remote tracking. What issues are you facing, which this functionality will help to resolve?

jennifer12121:
Hi @alberttorosyan! Thank you for getting back to me and apologies for the delay in my reply. 

Adding this feature will allow us to implement server interceptors that authenticate users using tokens and client interceptors to allow users to log in using tokens and attach the tokens to the gRPC requests to enhance security.

I have an implementation that I can contribute if that’s okay with you?

alberttorosyan:
Hi @jennifer12121!
We've recently reimplemented Aim remote tracking, and it now uses HTTP/ws to transfer data to remote.
As this issue is specific to gRPC, I think it's not relevant anymore?

PR #2459, opened 2023-01-05
[feat] Add bearer_token to grpc header

jennifer12121:
Related issue: https://github.com/aimhubio/aim/issues/2207


alberttorosyan (commented):
@jennifer12121 thanks a lot for contribution! Nice job 🙌

I'd suggest to change the implementation a bit to avoid the need of passing the `bearer_token` through multiple objects `__init__`s.
A couple of possible options:
- add a `classmethod` to the `class Client` to set token directly (without passing it to Repo/Run classes)
- read token value from pre-defined environment variable or file
- introduce a new Client config. accessible from any part of the code and set the token value as a config option.

jennifer12121:
Accidentally closed while updating to the latest version. Reopening.

jennifer12121:
Hi @alberttorosyan , apologies on the delay on this! I took your second suggestion to create a new environment variable called `AIM_RT_BEARER_TOKEN` and I only pass that into the `create_remote_channels` method inside of `Client`. Thank you for your feedback and please let me know if you have any questions/comments!

alberttorosyan:
Hey @jennifer12121!
Thanks for applying the changes and sorry for the delay. Will review the changes once again and get back to you 🙌

aim/ext/transport/client.py
```diff
@@ -76,15 +77,20 @@ def _create_remote_channels(self):
 
         ssl_certfile = os.getenv(AIM_CLIENT_SSL_CERTIFICATES_FILE)
         msg_max_size = int(os.getenv(AIM_RT_MAX_MESSAGE_SIZE, AIM_RT_DEFAULT_MAX_MESSAGE_SIZE))
+        bearer_token = os.getenv(AIM_RT_BEARER_TOKEN)
```
mihran113, line 80: Won't this just fail due to the fact that: 
```
AIM_RT_BEARER_TOKEN = None
```

aim/ext/transport/config.py
```diff
@@ -11,3 +11,4 @@
 # GRPC OPTIONS
 AIM_RT_MAX_MESSAGE_SIZE = '__AIM_RT_MAX_MESSAGE_SIZE__'
 AIM_RT_DEFAULT_MAX_MESSAGE_SIZE = 16 * 1024 * 1024  # 16MB
+AIM_RT_BEARER_TOKEN = None
```
mihran113, line 14: I guess this should look like something like this: 
```
AIM_RT_BEARER_TOKEN = '__AIM_RT_BEARER_TOKEN__'
```

aim/ext/transport/client.py
```diff
@@ -76,15 +77,20 @@ def _create_remote_channels(self):
...
         options = [
             ('grpc.max_send_message_length', msg_max_size),
             ('grpc.max_receive_message_length', msg_max_size)
         ]
 
         # open a channel with router
         if ssl_certfile:
```
mihran113, line 87: this check also needs to be changed, so if the `bearer_token` is provided then we should go with the case of secure channel as well

aim/ext/transport/client.py
```diff
@@ -76,15 +77,20 @@ def _create_remote_channels(self):
...
+            self._call_credentials = grpc.access_token_call_credentials(bearer_token)
```
mihran113, line 88: Let's add an explicit check that bearer token exists, and if so, then add the credentials:
```
if bearer_token:
    self._call_credentials = grpc.access_token_call_credentials(bearer_token)
    root_certificates = grpc.composite_channel_credentials(
         root_certificates, self._call_credentials
     )
```

mihran113 (changes requested):

aim/ext/transport/config.py
jennifer12121: Thank you for this feedback! I made the change you suggested.

Issue #2434, opened 2022-12-15
Aim getting stuck, IO error: While lock file


arturspeiser:
## 🐛 Bug

I run aim remotely. After a while it grinds down to a halt, so that I'm unable refresh any information. 
This error message is output roughly twice a second: 

> Failed to index Run '9c737223e903474b99ed1293'. Error: b'IO error: While lock file: /home/aspeiser/.aim/meta/chunks/9c737223e903474b99ed1293/LOCK: Resource temporarily unavailable'. 

After restarting the the repo/server it runs again fine for a while.

### Environment

- Aim Version v3.15.1
- Python version 3.9.2
- OS Debian

alberttorosyan:
@arturspeiser could you please list all the fils in `.aim` related to the Run?
```
find .aim -name "* 9c737223e903474b99ed1293*"
```

arturspeiser:
![image](https://user-images.githubusercontent.com/109271265/207877308-05431e88-fd46-4ab2-b040-adb9a6b5195e.png)

alberttorosyan:
@arturspeiser, one more thing. Is running without remote tracking an option for you? Just thinking about workarounds to not block your work till the issue is resolved.

arturspeiser:
Nah, my whole workflow is fully remote. I'm also not 100% sure that the slowdowns are actually caused by that error. The speed is generally very inconsistent for me.

alberttorosyan:
> Nah, my whole workflow is fully remote. I'm also not 100% sure that the slowdowns are actually caused by that error. The speed is generally very inconsistent for me.

@arturspeiser how many parallel runs do you have? `aim up` command does background optimization once it detects a finished run. So having many runs in progress might be the reason of performance diff. The storage should eventually become optimized if there are no active runs.

arturspeiser:
Yes definitely feels like having ongoing runs slows it down considerable, though I never do more then 10 or so in parallel. Also I only track scalars, no images or anything. I do however track some metrics at every single iteration. Would it help to reduce the update frequency?

PR #2468, opened 2023-01-09
[fix] Open container in read-only mode for indexing

alberttorosyan:


alberttorosyan:
resolves https://github.com/aimhubio/aim/issues/2434

PR #2484, opened 2023-01-19
[feat] Provide support for signaling run events via Remote Tracking

alberttorosyan:
- Extract the file management (polling, touching) from the `class CheckIn` into separate class.
- Change remote tracking handlers to use kwargs.
- Implementation of proxy class for file management.


alberttorosyan:
resolves #2485

mihran113 (approved):
Looks pretty good! Please don't forget to update the changelog as well :)

PR #2495, opened 2023-01-23
[feat] Add Prophet integration

grigoryan-davit:
This PR introduces integration with Prophet (https://github.com/facebook/prophet). Prophet follows the sklearn interface but doesn't have a callback mechanism and isn't trained iteratively, so the actual logging is a bit different from other integrations (see the example in examples/prophet_track.py.


aim/sdk/adapters/prophet.py
```diff
@@ -0,0 +1,76 @@
...
+            self._run.close()
+
+    def track_metrics(self, metrics: Dict[str, float]) -> None:
+        """
+        Since Prophet doesn't compute loss during training,
+        only hyperparameters are logged by default.
+        This method can be used to log any user-provied validation metrics.
+        """
+        for metric, value in metrics.items():
+            self._run.track(value, name=metric, context={"subset": "val"})
```
gorarakelyan, line 76: I guess it would be better to make it more flexible, i.e. enabling users to pass context as a track_metrics argument. Thoughts/objections?

aim/sdk/adapters/prophet.py
```diff
@@ -0,0 +1,76 @@
...
+        self._log_cout = log_cout
+        self.setup(prophet_model.__dict__)
+
+    @property
+    def experiment(self) -> Run:
+        if not self._run:
+            self.setup()
+        return self._run
+
+    def setup(self, hparams: Dict[str, Any]) -> None:
```
gorarakelyan, line 41: I wonder when the setup method is called? Should it be done manually?

aim/sdk/adapters/prophet.py
grigoryan-davit: Yeah makes sense. I originally implemented it like this because of the unusual training procedure of Prophet (and forecasting in general I guess, where you mostly care about the backtest), but I was having second thoughts. Will do today!

aim/sdk/adapters/prophet.py
grigoryan-davit: Nope, it's currently called in the constructor! I thought that would be the easiest way to handle things (requiring minimal input from the user), but please let me know if that's incongruent with the general Aim workflow.

tmynn (approved):
tested, lgtm

aim/sdk/adapters/prophet.py
```diff
@@ -0,0 +1,76 @@
...
+        Prophet doesn't have a callback system and isn't trained iteratively.
+        Thus, only the hyperparameters and user-provided metrics are logged.
+        """
+        self._repo_path = repo
+        self._experiment = experiment
+        self._system_tracking_interval = system_tracking_interval
+        self._log_system_params = log_system_params
+        self._run = None
+        self._run_hash = None
+        self._log_cout = log_cout
```
alberttorosyan, line 32: Seems the `self._log_cout` member is not used? Should it be removed?

alberttorosyan (commented):
Looks good 👍 Left one comment.
Also please add the feature in `CHANGELOG.md`. You can check-out details on the contribution process [here](https://github.com/aimhubio/aim/blob/main/CONTRIBUTING.md#contribution-process)

aim/sdk/adapters/prophet.py
grigoryan-davit: Hi Albert jan, yup you're right I'll remove the log_cout it in a sec. Will also add to the changelog.

grigoryan-davit: Hey Albert jan, just addressed your comments in the latest commits :)

mihran113 (approved):
Looks great!

alberttorosyan (approved):
lgtm

alberttorosyan:
resolves #2496

alberttorosyan:
@grigoryan-davit, seems all comments are addressed. Lets target to merge this PR today and include it in the upcoming release.

Issue #2503, opened 2023-01-26
Add the ability to easily copy run hash with a single click


VkoHov:
## 🚀 Feature

As a user, it would be very helpful to have the ability to easily copy the run hash with a single click. Currently, the process of copying the run hash requires multiple steps, such as right-clicking and selecting 'Copy' or manually highlighting and copying the text. This can be time-consuming and frustrating, especially when working with a large number of experiments.

### Motivation
Adding a single-click option to copy the run hash would significantly improve the user experience and make it easier to track and reference specific runs. It would also save time, as users would not have to manually copy the run hash each time they want to reference it. This feature would be particularly useful for users who frequently need to share run hashes with others, such as in a team or collaboration setting.

PR #2504, opened 2023-01-26
[feat] Add the ability to easily copy run hash with a single click

VkoHov:
Added copy icon near the run hash in the run page


VkoHov:
Resolves [#2503](https://github.com/aimhubio/aim/issues/2503)

roubkar (approved):

Issue #2249, opened 2022-10-06
Scaling up with UI Kit v2: Improving user experience and streamlining design


arsengit:
## 🚀 Feature

Develop a new UI Kit for better performance and scalability

### Motivation

As our project continues to grow, we've noticed that our current UI Kit using `Material UI` is becoming more difficult to maintain and customize. We believe that creating a new UI Kit using `Radix` UI and `Stitches` packages will provide better performance, scalability, and customization capabilities.

### Pitch

The new UI Kit v2 will be designed with performance, scalability, and customization in mind. We propose creating new UI Kit components using `Radix UI` and `Stitches` packages, starting with the most frequently used components, and gradually moving towards replacing existing Material UI components with UI kit v2 counterparts.

The UI Kit v2 will include the following components:

- [x] `Button` component
- [x] `IconButton` component
- [x] `Input` component
- [x] `Badge` component
- [x] `ListItem` component
- [x] `Popover` component
- [x] `Checkbox` component
- [x] `Text` component
- [x] `Switch` component
- [x] `QueryBadge` component
- [x] `Radio button` component
- [x] Polymorphic `Box` component
- [x] `Icon` component
- [x] `Toast` component
- [x] `Separator` component
- [x] `Tooltip` component
- [x] `Dialog` component
- [x] `Tree` component
- [x] `Select` component
- [x] `Slider` component
- [x] `ToggleButton` component
- [x] `Table` component
 
Will be also implemented `tabler-icons` as the icon pack for the new UI Kit and set up a `Storybook` for easily view and test the components

PR #2506, opened 2023-01-27
[feat] Add `Slider` component

arsengit:
Added the `Slider` component for ui kit v2


arsengit:
Resolves #2249

KaroMourad (commented):
Please separate "Slider" component styles with separate .style file, as it done for Tooltip component

aim/web/ui/src/components/kit_v2/Slider/Slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+    onValueChange,
+    ...props
+  }: ISliderProps) => {
+    const [sliderValue, setSliderValue] =
+      React.useState<number[]>(defaultValue);
+
+    React.useEffect(() => {
+      if (value) {
+        if (value && !_.isEqual(value, sliderValue)) {
+          setSliderValue(value);
```
KaroMourad, line 34: please remove first `if` condition, since `value` double checked

KaroMourad: Let's consider that `value` property could be `0` as well , and it's falsy value. 
And the condition will not pass if the value is equal to 0.

aim/web/ui/src/components/kit_v2/Slider/Slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <SliderRange className='SliderRange' />
+        </SliderTrack>
+        {sliderValue.map((_: number, i: number) => (
+          <SliderThumb className='SliderThumb' key={i}>
+            <SliderLabel className='SliderLabel'>{sliderValue[i]}</SliderLabel>
+          </SliderThumb>
+        ))}
+
+        {marks.length > 0 &&
+          marks?.map((mark) => {
```
KaroMourad, line 76: In case of incorrect `marks` value (undefined, null, etc...) this condition wouldn’t work and will break the app.

let's fix the condition: 
from: `marks.length > 0 && marks?.map((mark) => ...`
to: `marks?.length > 0 && marks.map((mark) => ...`

KaroMourad (approved):
lgtm!

PR #2518, opened 2023-02-01
[feat] Support multiple datasets tracking for HF trainer

tmynn:
Added support to track multiple datasets metrics in case one uses a dictionary of datasets for evaluation and/or training

### DIFF
--- a/aim/sdk/adapters/hugging_face.py
+++ b/aim/sdk/adapters/hugging_face.py
@@ -1,5 +1,6 @@
 from logging import getLogger
-from typing import Optional
+from typing import Optional, List, Dict
+from difflib import SequenceMatcher
 
 from aim.ext.resource.configs import DEFAULT_SYSTEM_TRACKING_INT
 from aim.sdk.num_utils import is_number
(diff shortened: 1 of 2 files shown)


mihran113 (approved):
Looks good, but let's just bring back the single quotes as we use single quotes everywhere else in the code

aim/sdk/adapters/hugging_face.py
```diff
@@ -108,15 +114,38 @@ def on_log(self, args, state, control,
...
 
     def close(self):
         if self._run:
             self._run.close()
             del self._run
             self._run = None
 
+    @staticmethod
+    def find_most_common_substring(names: List[str]) -> Dict[str, int]:
+        substring_counts: Dict[str, int] = {}
```
alberttorosyan, line 133: @tmynn I'm not sure I fully understand the code below.
As we're getting only the first key of the dict, does it make sense to fill in the rest of matches? 🤔

aim/sdk/adapters/hugging_face.py
```diff
@@ -97,6 +96,13 @@ def on_log(self, args, state, control,
                 if log_name.startswith(prefix):
                     log_name = log_name[len(prefix):]
                     context = {'subset': prefix[:-1]}
+                    if '_' in log_name:
+                        sub_dataset = AimCallback.find_most_common_substring(
```
alberttorosyan, line 100: lets restructure this code to make it more readable.

alberttorosyan (changes requested):
Overall looks good! Requested some changes before approving.

aim/sdk/adapters/hugging_face.py
tmynn: @alberttorosyan good point! Need to take the key with max value

aim/sdk/adapters/hugging_face.py
tmynn: Any recommendations on this?

gorarakelyan:
merging this to include in the upcoming release. let's improve code health afterwards

gorarakelyan (approved):

alberttorosyan (approved):
Approving changes. Code structure change will be applied later.

PR #2524, opened 2023-02-03
[fix] Fix empty-illustrations font-size for Figures and Audios explorers

KaroMourad:
-  Fixed empty-illustrations font-size for Figures and Audios explorers


KaroMourad:
Resolves #2525

arsengit (approved):

PR #2531, opened 2023-02-07
[fix] Use non-strict mode when saving HF model metadata

alberttorosyan:


mihran113 (approved):
Looks good, please link the issue as well

alberttorosyan:
resolves #2546

Issue #2526, opened 2023-02-06
Error loading non-builtins with Aim Logger and pl lightning with Aim server


charyeezy:
## 🐛 Bug

Problems uploading non builting obj in config when using Aim Logger and pl lightning when logging with Aim server.
However, when logging locally, the objects are stringified appropriately.

eg. with AIM Logger setup
```
aim_logger = AimLogger(
        repo="aim://my-aim-server:53800",
        experiment=experiment_name,
        train_metric_prefix="train/",
        val_metric_prefix="val/",
        test_metric_prefix="test/",
        # log_system_params=True,
    )
datamodule_args = {
        "root": ROOT_DATA_DIR,
        "train_transforms": get_train_transform(train=True),  
        "valid_transforms": get_valid_transform(train=False),
        "batch_size": batch_size,
        "num_workers": num_workers,
        "collate_fn": safe_collate_fn,
        "crop": False if args.raw else True,
    }
datamodule = DataModule(**datamodule_args)
trainer_args = {
        "default_root_dir": ROOT,
        "logger": aim_logger,
        "max_epochs": max_epochs,
        "log_every_n_steps": log_every_n_steps,
        "fast_dev_run": args.debug,
        "enable_checkpointing": True,
        "val_check_interval": 0.5,  ## since we have a big training set
        **accelerator_args,
    }
trainer = pl.Trainer(**trainer_args, auto_lr_find=True)
trainer.fit(model, datamodule=datamodule)

```

```zsh
...
Traceback (most recent call last):
  File "/home/coder/rosa/.venv/lib/python3.8/site-packages/pytorch_lightning/trainer/call.py", line 36, in _call_and_handle_interrupt
    return trainer.strategy.launcher.launch(trainer_fn, *args, trainer=trainer, **kwargs)
  File "/home/coder/rosa/.venv/lib/python3.8/site-packages/pytorch_lightning/strategies/launchers/subprocess_script.py", line 90, in launch
    return function(*args, **kwargs)
  File "/home/coder/rosa/.venv/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 645, in _fit_impl
    self._run(model, ckpt_path=self.ckpt_path)
  File "/home/coder/rosa/.venv/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1086, in _run
    self._log_hyperparams()
[...]

PR #2532, opened 2023-02-07
[fix] Add `set()` method implementation for RT TreeView classes

alberttorosyan:
The `ProxyTree` and `SubtreeView` classes do not provide `set()` method implementation, thus using the default `TreeView.set`. It falls-back to the `__setitem__`, which in turn ignores the strict flag.


alberttorosyan:
resolves #2526

mihran113 (approved):

Issue #2539, opened 2023-02-09
Aim telemetry hangs in air-gapped environment


wlhjason:
## 🐛 Bug

Within an environment with no internet access, the Aim SDK can hang on exit while Segment attempts to flush telemetry data to its servers.

### To reproduce

Create a `run_aim.py` script:

```python
import aim
import logging

logging.basicConfig()
run = aim.Run()
```

and run `python run_aim.py`.

This hangs on exit and produces log messages like this:

```
INFO:backoff:Backing off send_request(...) for 0.8s (requests.exceptions.ConnectTimeout: HTTPSConnectionPool(host='api.segment.io', port=443): Max retries exceeded with url: /v1/batch (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7fe251a0f190>, 'Connection to api.segment.io timed out. (connect timeout=15)')))
INFO:backoff:Backing off send_request(...) for 1.8s (requests.exceptions.ConnectTimeout: HTTPSConnectionPool(host='api.segment.io', port=443): Max retries exceeded with url: /v1/batch (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7fe251a2efa0>, 'Connection to api.segment.io timed out. (connect timeout=15)')))
INFO:backoff:Backing off send_request(...) for 1.4s (requests.exceptions.ConnectTimeout: HTTPSConnectionPool(host='api.segment.io', port=443): Max retries exceeded with url: /v1/batch (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7fe2519ba130>, 'Connection to api.segment.io timed out. (connect timeout=15)')))
```

### Environment

- Aim Version: **3.6.0** (following https://github.com/aimhubio/aim/pull/2490/)
- Python version: 3.8.13
- pip version: 22.2.2
- OS: Ubuntu 20.04.4

### Additional context

Manually disabling Aim telemetry beforehand via `$ aim telemetry off` (within the same Python environment) resolves the issue, but it would be great if this wasn't necessary!

SGevorg:
Thanks for sharing this @wlhjason. Looking into it. 
Adding @alberttorosyan and @mihran113 to this thread as well.

wlhjason:
Thanks for the quick response!

As a related issue, I noticed that `analytics.track_install_event()` (which is called at import time) calls out to Segment if it detects that Aim is running in a new Python environment. Thus:

[...]

PR #2542, opened 2023-02-14
[fix] Reduce segment flush() timeout interval and retry count

alberttorosyan:


alberttorosyan:
resolves #2539

aim/utils/tracking.py
```diff
@@ -68,7 +64,13 @@ def initialize(self) -> None:
         if self.initialized:
             return
 
+        segment_logger = logging.getLogger('segment')
+        segment_logger.disabled = True
```
mihran113, line 68: why do we disable the segment logger?

alberttorosyan: Even with the `logLevel == ERROR` segment might output errors. For example in the mentioned case `flush()` will output the timeout errors.
With the logLevel `INFO` the logs get flooded.

mihran113 (approved):

PR #2561, opened 2023-03-02
[feat] Add exception-free mode to Aim

alberttorosyan:
Introduce the safe mode into tracking with Aim.
- In order to mark the function as "safe-mode-enabled", decorate it with  `@noexcept`
- In order to enable/disable safe mode:
```python
from aim.utils import enable_safe_mode

enable_safe_mode()
...
```

_By default, in safe mode the raised exception and the function name will be logged with WARNING log level._

- In order to customize the exception handling need to register the callback on `TrainingFlow` instance
```python
class ExceptionHanlderCallback:
    @on.exception_raised
    def handle_exception(self, exception, function, **kwargs):
        # handling logic here

flow.register(ExceptionHanlderCallback())
```

and call

```python
flow.handle_exceptions()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 3.16.2
+- Add exception-free mode to Aim (alberttorosyan)
 - Expose `capture_terminal_logs` argument for `aim.sdk.adapters` classes (mihran113)
 
 ## 3.16.1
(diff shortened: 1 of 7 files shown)


mihran113 (approved):
looks great! Please add changelog entry as well. And also don't we need some docs for this?

alberttorosyan:
Resolves #2563

Issue #2552, opened 2023-02-28
Hiding all lines and then enabling a single line results in the wrong metric name.


froody:
## 🐛 Bug

Hiding all lines and then enabling a single line results in the wrong metric name.

### To reproduce

1. Select two runs in the runs table with multiple metrics (ideally the same set of metrics) 
2. Click "Compare" -> "Metrics" to open up the Metric Explorer
3. Click "Hide Lines" -> "Hide All Lines"
4. Select a metric near the bottom of the list and click "Show Lines"

Also note that the "Show Lines" button flips to "Hide Lines" after some time, this is also a bug.

### Expected behavior

A single plot for the selected metric with the correct name

### Environment

- Aim Version: 3.16.0
- Python version: 3.10.8
- pip version: 23
- OS (e.g., Linux): Ubuntu 22.04
- Any other relevant information: Using aimlflow to sync runs from mlflow

### Additional context

https://user-images.githubusercontent.com/241317/221999582-7adb0d86-38bf-4227-997c-4823264f24c3.mov

roubkar:
Hey @froody, thanks for reporting this issue, looking into it.

roubkar:
Hey @froody, the fix for this issue has been shipped.
To upgrade please run `pip install aim --upgrade`

SGevorg:
closing this issue as it was shipped.

PR #2567, opened 2023-03-05
[fix] Recover hidden lines state after the live update

roubkar:
- Fix the issue with hidden runs state reset after the live update
- Fix the issue with the wrong chart index being provided when there are hidden charts


roubkar:
Resolves https://github.com/aimhubio/aim/issues/2552

arsengit (approved):

Issue #2579, opened 2023-03-09
Add quick navigation to explorers from the run page


roubkar:
## 🚀 Feature

Add the ability to quickly navigate to explorer pages from the Run page

### Motivation

Provide the ability to easily use explorer features (grouping, visualization controls, etc...) for a single run. 

### Pitch

Add a popover (like the table compare popover in explorer pages) to the Run page with list items for navigating to each explorer.

### Alternatives

Alternative steps:
- Copy the run hash
- Go to the explorer page
- Filter runs based on run hash and search

PR #2580, opened 2023-03-09
[feat] Add quick navigation to explorers from the run page

roubkar:
- Added compare popover to the run page header
- Added `buttonText` property to compare popover
- Added missing explorer names to the compare selected runs config


roubkar:
Resolves https://github.com/aimhubio/aim/issues/2579

arsengit (approved):

Issue #2581, opened 2023-03-09
Ability to pass `run.experiment.name` and `run_hash` to the pl adapter


constd:
## 🚀 Feature

Add the ability pass a run name (in case of a new run) or a run hash (in case of an existing run) to the pytorch lightning logger.

### Motivation

Currently, using the AimLogger adapter for pytorch lightning makes it impossible to pass in a name for a run, resulting in a lot of runs names like `Run_<hash>` which make it hard to follow/track basic differences. At the same time, resuming a run, requires passing in a `run_hash` immediately after instantiation (and before `self.experiment` is accessed).

### Pitch

I would like to pass in run `name` and `hash` to the pytorch lightning aim logger, adding support for custom names and resuming training runs.

### Alternatives

```
logger = AimLogger()
logger._run_hash = <hash_number>
```
and
```
logger = AimLogger()
logger.experiment.name = <run_name>
```

alberttorosyan:
Hey @constd! Thanks for submitting the issue.
Putting the discord convo [link](https://discord.com/channels/1047782741736968202/1083237141531856937) for future reference.

PR #2585, opened 2023-03-11
[feat] Add run name & hash to lightning adapter args

constd:


constd:
@alberttorosyan & @mihran113 this is making it easier to pass in and set a custom name to a run, as well as pass a run_hash in order to support resuming training runs

aim/sdk/adapters/pytorch_lightning.py
```diff
@@ -35,17 +35,20 @@ class AimLogger(Logger):
     def __init__(self,
                  repo: Optional[str] = None,
                  experiment: Optional[str] = None,
+                 run: Optional[str] = None,
```
alberttorosyan, line 38: lets name the argument `run_name` to not be confused with the `Run` object.

alberttorosyan (commented):
Thanks for the contribution @constd! Left a small comment wrt argument naming.
Please also update the `CHANGELOG.md` 🙌

aim/sdk/adapters/pytorch_lightning.py
```diff
@@ -35,17 +35,20 @@ class AimLogger(Logger):
     def __init__(self,
                  repo: Optional[str] = None,
                  experiment: Optional[str] = None,
+                 run_name: Optional[str] = None,
```
alberttorosyan, line 38: Lets also move the argument to the back of the list. As the `AimLogger` allows both positional and keyword args, it might mess the order.

aim/sdk/adapters/pytorch_lightning.py
```diff
@@ -80,6 +83,7 @@ def experiment(self) -> Run:
                     system_tracking_interval=self._system_tracking_interval,
                     capture_terminal_logs=self._capture_terminal_logs
                 )
+                self._run.name = self._run_name
```
alberttorosyan, line 86: Just realized we need to check the `self._run_name` first. Passing value `None` actually means to leave the name as it was, so no need to assign in that case. Thoughts?

alberttorosyan (changes requested):
Went over changes once again. Found two things to change.

alberttorosyan:
@constd, tagging you just in case you missed my last comments 🙌

aim/sdk/adapters/pytorch_lightning.py
constd: Agreed! Good catch! :)

constd:
> @constd, tagging you just in case you missed my last comments 🙌

I did miss the comments @alberttorosyan -- just made the changes

mihran113 (approved):
@constd looks great! Thanks a lot for contribution.

mihran113:
Resolves #2581

alberttorosyan (approved):
LGTM

aim/sdk/adapters/pytorch_lightning.py
```diff
@@ -88,7 +93,6 @@ def experiment(self) -> Run:
                     log_system_params=self._log_system_params,
                     capture_terminal_logs=self._capture_terminal_logs,
                 )
-                self._run_hash = self._run.hash
```
mihran113, line 91: @constd just noticed this, I guess this shouldn't have been deleted, could you please bring it back, so I can merge the PR?

mihran113 (approved):

PR #2590, opened 2023-03-15
[Feat] Add -y option to Aim CLI commands

### LINKED ISSUE #2440: Introduce `-y` option in aim CLI commands
## 🚀 Feature

Aim commands such as `aim init` prompt user for inputs, which are mainly confirmation prompts. Need to add a command flag that will automatically confirm the action.

### Motivation
Enable aim commands in non-interactive environments (shell scripts, etc.)

### Pitch
Add -y option similar to apt install to automatically confirm the action.

emekaokoli19:
Resolves #2440 @alberttorosyan or @gorarakelyan please take a look at this

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.ts linguist-detectable=false
+*.tsx linguist-detectable=false
 *.scss linguist-detectable=false


gorarakelyan:
@emekaokoli19 thanks for the contribution! We will review it shortly and try to ship with the upcoming 3.16.3 version (planned for this week).

aim/cli/up/commands.py
```diff
@@ -75,7 +76,10 @@ def up(dev, host, port, workers, uds,
     if repo_status == RepoStatus.MISSING:
         init_repo = None
         if not force_init:
```
alberttorosyan, line 78: As we have a dedicated `force_init` flag, does it make sense to check the `--yes` here?

alberttorosyan (commented):
@emekaokoli19 great work! Thanks for your contribution. Left a comment related to the `aim up` command.

aim/cli/up/commands.py
emekaokoli19: @alberttorosyan sorry for the late reply, I didn't notice the `force_init` flag, will add a commit to fix this now

Issue #2615, opened 2023-03-24
Problem with scrolling to overflown content in Run page Texts tab


roubkar:
## 🐛 Bug

There's no way to see the entire text when the text cell content height exceeds 300px.

https://user-images.githubusercontent.com/9951653/227451542-e7b3a074-89e3-401c-bce6-79018ba5ac80.mov

### To reproduce

Track multiline text so that tthe text cell content height will be more than 300px

### Expected behavior

Text cell content should be scrollable

PR #2616, opened 2023-03-24
[fix] Add vertical scroll to Run Texts table cell

roubkar:
Fixed issue with scrolling to overflown content in the Run page Texts tab.

Added height to text cell and padding.


roubkar:
Resolves https://github.com/aimhubio/aim/issues/2615

KaroMourad (approved):
lgtm

PR #2618, opened 2023-03-24
[fix] Fix the UI crashing issue when opening the Metrics, Params, or Scatters Explorers

KaroMourad:
When opening Metrics, Params, or Scatters Explorers with empty chart data the UI crashes.

- Fixed the issue


arsengit (approved):

KaroMourad:
Resolves: #2619

Issue #2424, opened 2022-12-12
Performance issues when viewing images by selecting the images tab of run in aim ui


Dong-Ki-Lee:
## 🐛 Bug

After saving the image over several steps, if I try to check it in the aim ui, there will be a part that stops or becomes very slow while loading the image.

In the above situation (slow or frozen), all other requests are not received regardless of the number of workers.

### To reproduce
After saving the images on the remote server with the code below, the issue occurs when reading the image on the run page.

change [aim ip] and [model-path] in the code below
check this repo https://github.com/huggingface/diffusers
i tested with stable diffusion 2.0 model

```
import argparse
import hashlib
import itertools
import math
import os
from pathlib import Path
from typing import Optional

import torch
import torch.nn.functional as F
import torch.utils.checkpoint
from torch.utils.data import Dataset

from accelerate import Accelerator
from accelerate.logging import get_logger
from accelerate.utils import set_seed
from diffusers import AutoencoderKL, DDPMScheduler, StableDiffusionPipeline, UNet2DConditionModel, StableDiffusionImg2ImgPipeline
from diffusers.optimization import get_scheduler
from huggingface_hub import HfFolder, Repository, whoami
from PIL import Image
from torchvision import transforms
from tqdm.auto import tqdm
from transformers import AutoTokenizer, PretrainedConfig
from aim import Run, Image
import aim


logger = get_logger(__name__)
aim_run = Run(repo='[aim ip]', experiment='test', log_system_params=True)
aim_run.name = 'test'

save_path = '[model-path]'

p = StableDiffusionPipeline.from_pretrained(save_path, torch_dtype=torch.float16, revision='fp16').to('cuda')

prompt = [
    "test",
    "test",
    "test",
    "test",
    "test",
    "test",
    "test",
    "test",
    "test",
    "test",
    "test",
]

num_inference_steps = 25
guidance_scale = 9.0
num_samples = 5

for i in range(len(prompt)):
    images = p(
        prompt[i],
        num_images_per_prompt=num_samples, 
        num_inference_steps=num_inference_steps, 
        guidance_scale=guidance_scale,
    ).images

    aim_images = []
    
    for j in range(len(images)):
        aim_images.append(aim.Image(images[j], caption=f'{prompt[i]} examples'))
        
    aim_run.track(value=aim_images, name='generated images', step=i)
```

### Expected behavior

The image should be loaded and displayed without slowing down or stopping.

[...]

PR #2636, opened 2023-03-31
Bug Fix: A bug where images load slowly when accessing pages with many images. (in remote server. eg, aws)

asynclee:
About #2424 issue.

### Bug summary 

Bug where images load slowly when accessing pages with many images. Additionally, when the loading is slow, the backend does not receive any additional requests, which causes slow access as well.

### Why the bug is not detected in the local environment ?

In the local environment, the speed of reading images was so fast that the issue did not cause any bottleneck. However, in a remote environment, such as an EC2+EBS environment, the disk response speed is not as fast as a local disk, so the logic related to loading images can cause a bottleneck. This results in the issue where the backend does not receive additional requests when the loading is slow, leading to slow access.

### What part is the problem ?

In the existing logic, when receiving image uri arrays through the getbatch logic, the uri array is initialized after the loading is completed. In other words, for example,

Assume that 50 image uri arrays are being processed,
While processing the above images, if 20 uri array requests come in, a getbatch operation requesting 70 uri arrays including the 20 uri arrays additionally requested on the existing 50 uri array being processed is performed.
If this process is repeated, many duplicated images will be requested.

### How was it changed?

Changed the logic to additionally manage the uri arrays that are being processed/processed, and to avoid processing duplicate image uris that are already being processed.


roubkar:
Hey @asynclee, thanks for opening this PR 🙌 
I will review it shortly

roubkar:
Hey @asynclee, could you please also update the `CHANGELOG.md`?

asynclee:
Hello @roubkar. Sure. I'll update that.

roubkar:
> Hello, @roubkar. I wrote it based on what was written before, but I'm not sure if it's correct. Can you review it, please?

For now, I'm not sure whether this will be released with v3.17.3 or v3.18.0.
If you don't mind I will adjust that part later

roubkar (approved):
LGTM

roubkar:
Hey @asynclee, I just merged this PR.
Thanks for fixing this issue and congrats on your contribution 🎉

PR #2638, opened 2023-04-04
[feat] Expose context in the PyTorch Ignite

tmynn:
Developed a new feature that allows an initial context to be passed as an argument to the AimLogger


tmynn:
Resolves #2638

alberttorosyan (commented):
@tmynn, looks good!
One question I have is the following: what if the passed context already has a 'subset' key? 🤔 
Would it be reasonable to not overwrite its value in this case?

tmynn:
> @tmynn, looks good! One question I have is the following: what if the passed context already has a 'subset' key? 🤔 Would it be reasonable to not overwrite its value in this case?

Good point @alberttorosyan, should we rename the upcoming 'subset' key in that case?

mihran113:
> > @tmynn, looks good! One question I have is the following: what if the passed context already has a 'subset' key? 🤔 Would it be reasonable to not overwrite its value in this case?
> 
> Good point @alberttorosyan, should we rename the upcoming 'subset' key in that case?

I guess user's input must have higher priority, so if the `subset` is already present, we can leave it as it is, and don't override it afterwards

alberttorosyan (approved):
lgtm

mihran113 (approved):

PR #2640, opened 2023-04-06
[fix] Fix the community popup overflowing issue

KaroMourad:
When the sidebar items are too much and the screen is small, community popup overflows the UI body.

- Fixed the community popup overflowing issue


KaroMourad:
Resolves: #2641

roubkar (approved):

Issue #2432, opened 2022-12-15
"Reports" feature


rodrigo-castellon:
## 🚀 Feature

A reports feature that lets you document a set of related experiments together as a writeup with inline plots/graphs/tables from those experiments that you can look back on in perpetuity.

### Motivation

My workflow for a single "experiment" usually consists of (1) thinking of an idea / modification to my model, (2) running with and without this modification in parallel, then (3) comparing these two runs under different metrics I care about. Then, after I've moved on and gone through this process many times, I lose track of exactly which experiments I've run, and can only see a raw list of all of my runs, which loosely correspond to many different experiments. This is a problem because after I've worked on a project for a while, I may be interested in looking back to remind myself how a particular experiment went from early in the project, but I am only able to at best guess which were exactly the runs corresponding to this.

### Pitch

W&B, which is the platform I used before Aim, provides "Reports", which allows you to create a report for each experiment you run, so that each report tells you what the hypothesis was and accompanying results. It would be nice to get a feature very similar to W&B, that allows me to store my experimental hypotheses + results in atomic units for safekeeping in perpetuity.

Specifically, this could come in the form of a new tab on the left-hand side of the Aim UI that says "Reports". In this tab, you could create a new report, and it would give you access to an editor where you could create a title, add a text writeup and insert plots/graphs/tables inline into that writeup. Said plots/graphs/tables are of course linked to actual runs. When you save the report and view it later, you can click on the plot/graph/table and view the runs that produced those results. When you go back to the "Reports" tab, it will have a list of your reports so far, and you can keep creating new ones or edit old ones.

### Alternatives

Potential alternatives are to keep said reports in a spreadsheet / google doc separately, which is what I have been doing, but it would be nice for this to be integrated into the platform.

### Additional context

See W&B reports here: https://wandb.ai/site/reports.

gorarakelyan:
@rodrigo-castellon thanks for the detailed suggestion! We do have this in the roadmap. But currently the core contributors are focused on improving existing overall experience, as well as the tech core (performance, stability, etc). We plan to add reports-like feature after that (probably in 1-2 months). Will keep you updated on the progress.

rodrigo-castellon:
Thank you! Looking forward to the improvements.

LangDaoAI:
any progress on the feature?

roubkar:
> any progress on the feature?

This feature is currently in an early development stage, will share progress under this issue.

roubkar:
[...]

PR #2676, opened 2023-04-24
[feat] Add the ability to create Reports

roubkar:
With these changes, users will be able to create reports with plain markdown using custom board code to embed visualizations in their markdowns.

- Added Reports list page
- Added Report page with edit mode
- Added `react-markdown` library for rendering md as HTML
- Add logic for replacing Aim Board code block with visualizations


roubkar:
Resolves https://github.com/aimhubio/aim/issues/2432

Issue #2689, opened 2023-04-27
Core Python SDK allowing extensibility and costumization


alberttorosyan:
## 🚀 Feature

Aim v3.x SDK provides classes for tracking experiments metadata in a form of `Run` objects. A pre-defined set of methods and metadata types allows covering ML experiment tracking in wide variety of cases.
However, the `Run` object comes with pre-set capabilities and extending it is not straight-forward and easy task.
Additionally, Aim SDK provides pythonic query API to query and collect tracked Runs, Metrics and object sequences. This interface is also limited to the pre-defined set of metadata types, and has some usability and ergonomics issues.

Need to design and implement new SDK, which will address these issues, mainly:
1. Provide easy and intuitive mechanism to define new metadata types, sequences and collections.
2. Provide low-level interface, allowing to iterate over and collet tracked metadata.
3. Provide query interface layer, allowing to query tracked data in a declarative manner, based on the data type, value and additional metadata.

### Motivation

Address UX issues present in Aim v3.x SDK, provide SDK with extensibility features in place.

Issue #2556, opened 2023-03-01
Add the ability to create custom UI boards 🔥


roubkar:
## 🚀 Feature

Add the ability to create custom UI boards with Python

### Motivation

Currently, Aim UI offers a variety of built-in explorers, which are helpful for exploring and understanding multidimensional data.
However, often there is a need to create more customized and specialized views that are not available in the tool.

By enabling users to create custom boards with Python, they can leverage the power and flexibility of already familiar programming language to develop views that suit their specific needs which can vary based on the use case. 
For example, they could use Plotly to create interactive visualizations that enable exploring and visualizing data in new ways.

With this new feature users should be able to create boards like sweeps with ~50 lines of code.

### Pitch

- [x] Integrate Pyodide to be able to run python within the browser.
- [x] Add functionality for creating, updating, and deleting boards
- [x] Add pythonic API for fetching data, visualizing with built-in components and Plotly figures, and customizing the layout.
- [x] Add Boards page
- [x] Add Board page
- [x] Add edit mode with a built-in code editor

qifeng22:
How to use this function?
Add functionality for creating, updating, and deleting boards
@roubkar

say4n:
Hi, super excited about this -- any update on if this is available in the UI yet or plans of as such?

PR #2692, opened 2023-04-27
[feat] Update data fetch query endpoint path for boards

roubkar:
- Updated fetch query endpoint path 
- Added classes for built-in data types with filter method
- Removed async/await from data fetch methods
- Implemented querying functionality with an approach inspired by React suspense and Vercel swr

Once the search is called the query is done asynchronously and the execution continues taking an empty list as a result.
Once the query is fulfilled the python script is re-executed taking the query result from the cache.


roubkar:
Resolves https://github.com/aimhubio/aim/issues/2689, https://github.com/aimhubio/aim/issues/2556

KaroMourad (approved):

PR #2696, opened 2023-04-28
[fix] Notebook extension start-up

mihran113:


aim/cli/manager/manager.py
```diff
@@ -45,7 +45,7 @@ def run_up(args):
         'host': 'http://' + args['--host']
     }
 
-    time.sleep(2)
+    time.sleep(5)
```
alberttorosyan, line 48: how stable is this? 🤔
is there a robust way to make sure the aim up is running?

mihran113: Changed to `/status` api checks instead, I guess this one is more trustworthy

alberttorosyan (approved):
lgtm

PR #2698, opened 2023-04-30
[feat] Add python interface for RunLogs

roubkar:
- Add RunLogs to viz elements map
- Add python interface for RunLogs


roubkar:
Resolves https://github.com/aimhubio/aim/issues/2556

KaroMourad (approved):

PR #2704, opened 2023-05-02
Fix searching on <Enter> on Windows

### LINKED ISSUE #2703: Search is not triggered on <Enter> on Windows browsers
## 🐛 Bug

With any browser on Windows, when typing `Enter` in the search box, a newline is inserted instead of triggering the search.

### To reproduce

On a Windows browser, type a search expression in any page that supports it (Runs, Metrics, etc.), then type <Enter>.

### Expected behavior

The search gets triggered.

### Environment

- Aim Version: 3.17.3
- Python version: 3.10.10
- pip version: 22.3.1
- OS: Windows

jgiannuzzi:
Windows browsers send `\r\n` instead of just `\n` when `Enter` is pressed in the search box. This PR changes the way we detect this keypress so that searching can be triggered in the same way on Windows as on other platforms.

### DIFF
--- a/aim/web/ui/src/components/AutocompleteInput/AutocompleteInput.tsx
+++ b/aim/web/ui/src/components/AutocompleteInput/AutocompleteInput.tsx
@@ -186,7 +186,7 @@ function AutocompleteInput({
           //   : formattedValue;
           onChange(formattedValue, ev);
         }
-        if (ev.changes[0].text === '\n') {
+        if (/^\r?\n$/.test(ev.changes[0].text)) {
           formattedValue = hasSelection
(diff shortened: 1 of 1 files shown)


roubkar:
Resolves #2703

roubkar:
Hey @jgiannuzzi, I have modified the PR description so that the issue won't be automatically closed when we will merge the PR.

We usually close issues after releasing the changes.

jgiannuzzi:
Sounds good, thanks @roubkar!

arsengit:
Thanks, @jgiannuzzi, for the valuable fix!

Issue #2702, opened 2023-05-02
Dashboard shows log messages multiple times


Robert27:
## 🐛 Bug

As soon as you log a message and view it on the dashboard, the newest message is duplicated every second.
When you refresh the page, it is gone and starts over again.

### To reproduce

1. Log a message using log_debug/info etc.
2.Vview it on the dashboard.

### Expected behavior

Displaying the log messages without the new message being displayed repeatedly.

### Environment

- Aim Version: 3.17.3
- Python version 3.9.16
- pip version 23.1.1
- OS MacOS 13.3.1
- Browsers: Safari / Firefox

### Additional context

https://user-images.githubusercontent.com/53063597/235601527-1589c919-b958-44a6-b815-7cd99f7a4f98.mov

roubkar:
Hey @Robert27, thanks for reporting this, looking into it.

Robert27:
Thanks for the quick fix :)

PR #2714, opened 2023-05-04
[fix] Correct log records count calculation

roubkar:
Send log records count instead of last item index


roubkar:
Resolves https://github.com/aimhubio/aim/issues/2702

Issue #2718, opened 2023-05-04
Tensorboard histograms are not synced with Aim


alansaul:
## 🚀 Feature

Allow `aim.ext.tensorboard_tracker.tracker.TensorboardFolderTracker` to track `histograms` within tensorboard as an Aim `Distribution`.

### Motivation

At the moment `aim.ext.tensorboard_tracker.tracker.TensorboardFolderTracker` can sync `images` and `scalars`, and `audio` from tensorboard into Aim (though the plugin is not supported), but it currently is not able to sync `histograms`.

Histograms however do appear to have a direct metric type counterpart in Aim, namely the `Distribution` object, so there doesn't appear to be any extra backend work needed to support them, is my understanding.

Some tools that use tensorboard track `histograms` automatically, so it would be great to have these also recorded in Aim automatically.

### Pitch

Add histogram plugin and non-plugin support to `aim.ext.tensorboard_tracker.tracker.TensorboardFolderTracker`. 

### Alternatives

None

### Additional context

None

PR #2719, opened 2023-05-04
[feat] Add ability for `TensorboardFolderTracker` to track Tensorboard `Histogram`'s as Aim `Distribution`'s

alansaul:
Added an implementation to also allow Tensorboard `histogram`'s to be tracked as Aim `Distribution`'s.

I have only user-tested the plugin based version, but have added unit-tests following a similar testing methodology as #2669 (using torch's utilities for tensorboard events) for both the tensorboard histogram plugin and non-plugin implementations. 

- Add `histogram` -> `Distribution` conversion
- Add `histogram` plugin -> `Distribution` conversion
- Add unit-tests

Resolves: #2718


mihran113:
Hey @alansaul! Looks great to me. Will merge to main, then manually transfer to the `release/3.17.x` branch alongside the previous commits that were adding features to the Tensorboard tracker to include in the upcoming patch.

mihran113 (approved):

PR #2726, opened 2023-05-05
[feat] Optimize board rendering via updated layout blocks

roubkar:
- Set the `parent_block` field on elements during initial rendering
- Bumped up pyodide version to `0.23.2`
- Updated new board example code with the new API
- Fixed run messages with extra args height calculation


KaroMourad (approved):
lgtm

PR #2751, opened 2023-05-16
[feat] Add python interface for form block element

roubkar:
These changes enable users to update the board state (re-render) on submit action.
e.g. Search metrics for a given query (typed in a text input) after clicking the search button.

Changes:
- Added `form` block element
- Batch  component state updates if the parent block is a form element
- Fixed console panel style issues


KaroMourad (approved):
lgtm

PR #2784, opened 2023-05-25
[feat] Enable range slicing via sequence filter

roubkar:
Add ability to apply range slicing during data filtration in boards.

e.g.

```python
from asp import Metric

Metric.filter(query='True', count=10, start=100, stop=200)
```


gorarakelyan (approved):
lgtm

mihran113 (approved):

PR #2799, opened 2023-05-31
[refactor] Refactor object depth processing for non-metric explorers

KaroMourad:
- Refactored object depth processing for non-metric explorers


roubkar (approved):
👍

PR #2852, opened 2023-06-19
[feat] Add base path for aim tracking server

mihran113:


src/aimcore/transport/config.py
```diff
@@ -2,13 +2,7 @@
 AIM_SERVER_DEFAULT_PORT = 53800
 AIM_SERVER_DEFAULT_HOST = '0.0.0.0'
 AIM_SERVER_MOUNTED_REPO_PATH = '__AIM_SERVER_MOUNT_REPO_PATH__'
+AIM_SERVER_BASE_PATH = '__AIM_UI_BASE_PATH__'
```
alberttorosyan, line 5: should we use the same base path for both UI and TS?

mihran113: I don't think that's a good idea before unification, this was just a typo :)

alberttorosyan (approved):
lgtm

PR #2877, opened 2023-06-27
[feat] Extend Runs board functionality in ASP

roubkar:
- Added additional controls for Runs table
- Removed dummy board files
- Fixed server functions cache issue


alberttorosyan (approved):
LGTM

Issue #2907, opened 2023-07-11
No such file or directory: '~/.aim_profile'


tmynn:
## 🐛 Bug

When running a slurm job, the latter tries to access the file '~/.aim_profile' without having the necessary permission, and it fails with no such file error.
The exception is thrown on the following: https://github.com/aimhubio/aim/blob/c7f6bbc28500433e29cc958f92ad53cf6b900f60/src/python/aim/_ext/tracking/__init__.py#L12

### To reproduce

Run any slurm job.

### Expected behavior

To skip the logic if the permission for the file is not granted.

### Environment

- Aim Version: 3.17.3
- Python version: 3.9.16

PR #2922, opened 2023-07-14
[fix] Remove telemetry tracking

mihran113:


mihran113:
Resolves #2907

alberttorosyan (approved):
LGTM

One Q: would it make sense to also remove segment API calls from UI, instead of setting telemetry_enabled flag to 0? CC @roubkar

PR #2933, opened 2023-07-19
[feat] Add CLI to migrate Aim repos

alberttorosyan:


mihran113 (approved):
Overall looks good! Just maybe there's a need to rename the util functions for the sake of readability.

alberttorosyan:
> Overall looks good! Just maybe there's a need to rename the util functions for the sake of readability.

Renamed utility function to indicate migration from v3 to v4 format. Also merged to single run migration into the main utility function.

PR #2946, opened 2023-07-26
[feat] Add methods for sequence and container delete

mihran113:


src/python/aim/_sdk/collections.py
```diff
@@ -131,6 +141,23 @@ def count(self):
         # more optimal implementation
         return sum(1 for _ in self.__iter_meta__())
 
+    def delete(self):
+        from aim._sdk import Repo
+        from aim._sdk import Container
+        from aim._sdk.lock_manager import RunLockingError
+        repo = Repo.active_repo()
```
alberttorosyan, line 148: If we have multiple open repos, this line would fail. The `repo` instance should be available in the collecrtion's `query_context` dict.

src/python/aim/_sdk/sequence.py
```diff
@@ -301,6 +301,13 @@ def _check(self, query, query_cache, *, aliases=()) -> bool:
         query_params.update({cp: c_proxy for cp in container_alias_names})
         return query.check(**query_params)
 
+    def delete(self):
+        del self._data_loader()[(self._ctx_idx, self._name)]
+        del self._container_tree[(KeyNames.SEQUENCES, self._ctx_idx, self._name)]
+
+        self._info.empty = True
+        self._info.next_step = 0
+
```
alberttorosyan, line 310: Suggestion: set a flag which will prevent tracking on the deleted Sequence. Technically, the Python object still lives, and the track() function call is possible.

src/python/aim/_sdk/container.py
```diff
@@ -215,6 +215,36 @@ def _check(self, query, query_cache, *, aliases=()) -> bool:
...
+        # remove container blobs trees
+        blobs_tree = self.storage.tree(self.hash, 'BLOBS', read_only=False)
+        del blobs_tree.subtree(('meta', 'chunks'))[self.hash]
+        del blobs_tree.subtree(('seqs', 'chunks'))[self.hash]
+
+        # delete entry from container map
+        del meta_tree.subtree('cont_types_map')[self.hash]
+
+        # close the container
+        self.close()
```
alberttorosyan, line 246: `close()` will try to set container end_time. We will need to adjust that part to have just trees cleanup and resource release mode.

src/python/aim/_sdk/sequence.py
mihran113: I think there's no need to prevent tracking on a deleted sequence. In a deleted state it'll be empty, and tracking will be done from a step 0, the same as for newly initialized sequences

alberttorosyan (approved):
LGTM

PR #2961, opened 2023-08-22
[feat] Register `find()` method on packages pyodide module

roubkar:
- Added API calls for `find-sequence` and `find-container` endpoints
- Added `find` method mock on packages type classes for pyodide
- Update `run.py` board in `asp` package with usage of `Run.find()` method


alberttorosyan (approved):
LGTM

PR #2963, opened 2023-08-23
[fix] Add fallback json rendering for None value

roubkar:
When passing `None` to `ui.json` a `Something went wrong` error is being thrown.

Use fallback monospace text rendering to prevent the board page from crashing.


alberttorosyan (approved):
LGTM

PR #2964, opened 2023-08-24
[feat] Update Aim ML adapters

alberttorosyan:


pkgs/aimstack/ml/adapters/keras.py
```diff
@@ -84,7 +67,6 @@ def metrics(
 
     def close(self) -> None:
         if self._run:
-            self._run.close()
```
tmynn, line 87: Why this line is removed? Will it be enough to just delet or set the run instance to None?

tmynn (changes requested):
Seems like we need to update all docstrings, e.g. remove `capture_terminal_logs`.

Shall we add these two blocks inside of the `Run`,  they seem to be appearing in every adapter.

```
if self.experiment_name is not None:
    self._run.experiment = self.experiment_name
if self.log_system_params:
    self._run.enable_system_monitoring()
```

The rest lgtm!

tmynn (approved):

PR #2967, opened 2023-08-24
[feat] Enable form submissions with enter key

roubkar:
- Use `form` tag for form viz element in boards
- Add `onSubmit` action handler
- Fix types for `Button` component props


alberttorosyan (approved):
LGTM

PR #2976, opened 2023-09-07
[feat] Enable resending requests ignoring cache

roubkar:
- Signals are being used to invalidate query results cache on specific state updates
- By binding queries and components with the same signal you can control when to invalidate query results cache
- You can use signals as primitive strings or as Signal class instances
- With Signal class, you can also specify properties that should be present in state update to invalidate query results cache

### DIFF
--- a/src/aimcore/web/ui/public/aim_ui_core.py
+++ b/src/aimcore/web/ui/public/aim_ui_core.py
@@ -76,7 +76,7 @@ def dispatch_signal(signal_name=None, properties=None):
     if signal_name is not None:
         signal = signals_store.get(signal_name, None)
         signal_properties = signal.get("properties", None)
-        print('-----', properties, signal_properties)
+
         if (signal is None) or (
             properties is not None and 
             signal_properties is not None and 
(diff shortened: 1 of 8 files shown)


src/aimcore/web/ui/public/aim_ui_core.py
```diff
@@ -48,6 +48,56 @@ def wrapper(*args, **kwargs):
...
+            signals_store[signal]["query_keys"].append(query_key)
+
+def dispatch_signal(signal_name=None, properties=None):
+    """Dispatches signal. Cleares query results cache for all query keys bound to signal"""
+
+    if signal_name is not None:
+        signal = signals_store.get(signal_name, None)
+        signal_properties = signal.get("properties", None)
+
+        if (signal is None) or (
```
alberttorosyan, line 89: `signal is None` check should be done before getting its properties.

alberttorosyan (commented):
looks good! left a couple of comments

alberttorosyan (approved):
LGTM

PR #2979, opened 2023-09-08
[feat] Update query result processing for boards

roubkar:
To keep consistency between SDK retrieving methods return data format and UI retrieving methods return data format:
- Removed data processing from search and find methods used for `.filter` and `.find` methods in boards
- Updated boards' code based on the data format change

### DIFF
--- a/src/aimcore/web/ui/public/aim_ui_core.py
+++ b/src/aimcore/web/ui/public/aim_ui_core.py
@@ -171,11 +171,11 @@ def query_filter(type_, query="", count=None, start=None, stop=None, is_sequence
         if data is None:
             raise WaitForQueryError()
 
-        data = json.loads(data, object_hook=process_properties)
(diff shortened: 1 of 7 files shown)


src/aimcore/web/ui/public/aim_ui_core.py
```diff
@@ -144,7 +174,8 @@ def query_filter(type_, query="", count=None, start=None, stop=None, is_sequence
         data = json.loads(data, object_hook=process_properties)
 
         query_results_cache[query_key] = data
-        return data
+        
+        return ItemList(data)
```
alberttorosyan, line 178: shouldn't we return `ItemList` for cached results too?

alberttorosyan (changes requested):
Left one comment w.r.t. iterators and cached response results

alberttorosyan (approved):
LGTM

PR #2986, opened 2023-09-12
[feat] Allow multiple Aim packages in UI

alberttorosyan:


src/aimcore/web/api/projects/views.py
```diff
@@ -115,7 +89,7 @@ async def project_container_types_api(only_tracked: Optional[bool] = False):
 
 
 @projects_router.get('/actions/')
-async def project_actions_api():
+async def project_actions_api(packages=Depends(load_active_packages)):
```
mihran113, line 92: Why do we need to load all the active packages on every request?

mihran113 (approved):
left a minor comment, the rest looks good

src/aimcore/web/api/projects/views.py
alberttorosyan: we need to make sure that all the types/boards/actions info is loaded. each package will be imported only once and there's a ttl on reading the `active_pkg` file contents, so it should not have performance implications.

PR #2991, opened 2023-09-16
[fix] Keep state in URL for edit mode

roubkar:
- Rename `Edit` to `Playground` in boards
- Keep board state in URL while switching to the playground
- Stay in the playground while navigating through board links


KaroMourad (approved):
lgtm

PR #3003, opened 2023-09-27
[fix] Align row items by start in boards

roubkar:
Fix minor style issues as well:
- `cursor: pointer;` for table cells
- loading bar width calculation


alberttorosyan (approved):
LGTM

PR #3044, opened 2023-10-12
[fix] Fix typos in stat.py to collect gpu memory and power correctly

ChanderG:
typo in code led to these metrics not being collected at all


alberttorosyan (approved):
Looks great! Thanks for contribution @ChanderG 🙌

PR #3064, opened 2023-11-15
[fix] Update `upgrade` command for logging repo

tejank10:
`aim upgrade <repo_path> 2to3` command is now `aim storage <repo_path> upgrade 2to3`[1]. Updated the `repo_utils.py` to reflect the same.

References:
[1] https://aimstack.readthedocs.io/en/latest/refs/cli.html#storage


mihran113:
Hey @tejank10! Thanks a lot for noticing the mistake there, and sorry for late response. As of now we've decided to remove the `2to3` upgrade command, so this part of the code is gone now :)

PR #3083, opened 2024-01-04
[feat] Add refresh token mechanism on aim UI

KaroMourad:
- Added refresh token mechanism on aim UI


alberttorosyan (commented):
@KaroMourad the changes look good.
Please make sure to update design docs before merging. How the authentication works with Aim vs. AimHub backend, etc.

Issue #2999, opened 2023-09-22
Bug with aim.Run when using hash to locate a run


HoBeedzc:
## 🐛 Bug

I encountered a bug while using AIM. When attempting to utilize the `aim.Run` function to locate a run using its hash, I encountered the following error:

```Python
>>> aim.Run(run_hash='51031438759943878c6f9808')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/ext/exception_resistant.py", line 70, in wrapper
    _SafeModeConfig.exception_callback(e, func)
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/ext/exception_resistant.py", line 47, in reraise_exception
    raise e
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/ext/exception_resistant.py", line 68, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/sdk/run.py", line 828, in __init__
    super().__init__(run_hash, repo=repo, read_only=read_only, experiment=experiment, force_resume=force_resume)
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/sdk/run.py", line 276, in __init__
    super().__init__(run_hash, repo=repo, read_only=read_only, force_resume=force_resume)
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/sdk/base_run.py", line 50, in __init__
    self._lock.lock(force=force_resume)
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/storage/lock_proxy.py", line 38, in lock
    return self._rpc_client.run_instruction(self._hash, self._handler, 'lock', (force,))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/me/miniconda3/lib/python3.11/site-packages/aim/ext/transport/client.py", line 260, in run_instruction
    return self._run_read_instructions(queue_id, resource, method, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3090, opened 2024-01-11
[fix] Force reacquire lock in torch lightning

inc0:
When running testing on a Run, this line failed with error similar to #2999 . It was quite hard to debug, but it seems this is the culprit. It appears that Run was already locked and it failed with traceback when trying to run pytorch lightning training+testing run

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Add support for `sqlalchemy 2.0` (mihran113)
 - Add `min/max/first` values tracking for metrics (mihran113)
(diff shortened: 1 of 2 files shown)


alberttorosyan (approved):
LGTM

inc0:
Thanks @alberttorosyan  for review! Can you take a look at the CI failure please? It doesn't seem to be related to the PR.

alberttorosyan:
> Thanks @alberttorosyan for review! Can you take a look at the CI failure please? It doesn't seem to be related to the PR.

Yes, the unit-tests passed after re-run. Thank you for your contribution 🙌

PR #3091, opened 2024-01-17
[feat] Revamp the design of creating app models

KaroMourad:
- Split app model methods into separate files


alberttorosyan (approved):
LGTM

Issue #1971, opened 2022-07-18
Show max instead of last value in runs


Pyrestone:
## 🚀 Feature
In the runs explorer, I would like an option to show the max/min achieved value of a metric in the run list instead of only seeing the last value.

Ideally I would also like to sort by those.

### Motivation

I would like to know the best validation set accuracy of all my models without having to check each time series individually.
Ideally, I would also like to be able to sort by this.

### Pitch

click the 3 dots at the top of a column on the runs -> submenu opens, where I would like to have options "show min", "show max", "show first", "show last", and ideally an option "sort by this column"

### Alternatives

selection via "manage columns" would also be possible, so I could see last and max simultaneously.

### Additional context
![image](https://user-images.githubusercontent.com/20396757/179472986-2c2d27cf-70e8-42e0-a925-c79f36a3b9bf.png)

Add options 
- "show min"
- "show max"
- "show first"
- "show last"
- "sort by column" (switch ascending/descending)

to this menu:
![image](https://user-images.githubusercontent.com/20396757/179473104-104a53ae-8d86-4965-b691-4588ac94aaca.png)

Pyrestone:
@gorarakelyan Is this still on your roadmap? I would still find this feature very useful

gorarakelyan:
@Pyrestone sorry for a late reply.

It has been tough weeks in terms of number of critical issues and enhancements that we tried to build and ship for the community. Sometimes we are forced to prioritize even between critical issues. 

Currently we are really packed with issues for this and for the next week. The next week particularly is occupied, as we are planning to release Aim v3.14.0 (with revamped home page!), which requires both pre-release and post-release procedures.

Thus, I can not tell for sure if it will be possible to take over this issue within the next 10 days. Tough realization :( 
But, I totally agree on the importance and value of this feature request. It is for sure on the roadmap. It is only the matter of time and capacity.

We would do our best to build and ship it as a patch release of 3.14 version.

Hope this makes sense to you.

Thanks for your help and contribution to the community!

Pyrestone:
Hi @gorarakelyan, I understand that it's more important to focus on usability breaking issues first, (such as #2188 i presume?)
It's more important to me as well to have a stable, working, and portable version of aim than having fancy features.

I'm thankful for all the great work you and the team are doing on aim :pray:, keep going!

gorarakelyan:
Thanks for the kind words and understanding @Pyrestone. 

[...]

PR #3094, opened 2024-01-18
[feat] Add the ability to view metric min/max/first values in Runs Explorer

KaroMourad:
-  Added dropdown to select which metric value (min/max/first/last) you want to see in Runs table


KaroMourad:
Resolve: #1971

mihran113 (approved):
LGTM

Pyrestone:
Thanks @KaroMourad  for implementing this!
As I already mentioned in #1971 though, this feature only really becomes useful if we can select min/max/first/last on a per-column-basis, as it was initially described in the feature request. 

This is because some metrics are minimum-focused, such as loss, and others such as accuracy are maximum-focused, and for others still, such as step counters only the last value is important.
It would be really useful to be able to visualize all of this at the same time instead of alternating between four different views to gather this information.

@KaroMourad Do you think you could modify the implementation such that this becomes possible?

PR #3108, opened 2024-02-29
[fix] Serve files linked into the static files directory

martenlienen:
In certain conda environments, the files in the static files root directory can be symlinked from a different place. The fixed implementation only resolves relative .. segments in the request path without resolving any symlinks. This way, it still prevents reading arbitrary files through the web server while allowing the reading of symlinked files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # Changelog
 
-## 3.19.0 
+## 3.19.0
 
-### Enhancements: 
+### Enhancements:
 - Replace grpc with http/ws as transport for aim tracking server (mihran113)
 - Remove `aim storage upgrade 2to3` command (mihran113)
+
+### Fixes
+- Allow the web UI to serve assets symlinked into the static files directory (martenlienen)
 
 ## 3.18.1 Feb 7, 2024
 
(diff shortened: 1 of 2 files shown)


alberttorosyan (approved):
Looks good! Thanks for the contribution @martenlienen 🚀 
Waiting for @mihran113's review.

martenlienen:
I force-pushed to fix the style issue.

mihran113 (approved):
Looks good to me! @martenlienen can I ask you to add an entry in `CHANGELOG.md` file

martenlienen:
Done!

geraldino2:
Hey, I think this PR actually resulted in a 404 for all static files served from `static_files_root`. I guess the condition below should use `!=`. 

https://github.com/aimhubio/aim/blob/main/aim/web/api/views.py#L24-L25
   ```python3
 if common_prefix == static_files_root:
       raise HTTPException(status_code=404)
```

However, note that it would make the code vulnerable to a trivial path traversal: `curl http://127.0.0.1:43800/static-files/x/../../../../../../../../../../../../../secret_gzfile --path-as-is`.

**Comment by mihran113** [comment]
> Hey, I think this PR actually resulted in a 404 for all static files served from `static_files_root`. I guess the condition below should use `!=`.
> 
> https://github.com/aimhubio/aim/blob/main/aim/web/api/views.py#L24-L25
> 
> ```python
> if common_prefix == static_files_root:
>     raise HTTPException(status_code=404)
> ```
> 
> However, note that it would make the code vulnerable to a trivial path traversal: `curl http://127.0.0.1:43800/static-files/x/../../../../../../../../../../../../../secret_gzfile --path-as-is`.

Hey @geraldino2! Thanks a lot for noticing the wrong condition! 
Regarding the path traversal part: How would that be? 
```
    static_file_name = os.path.normpath(static_files_root / path)
```
this line of the code will resolver any `..`s in the path, and if the resulting path is outside of the build directory wouldn't the common prefix be different than the build directory itself?

geraldino2:
Hey @mihran113, you are right about path traversal. I didn't really read the code and I guess I may have removed the condition when debugging it. Sorry about that!

PR #3112, opened 2024-03-09
[fix] Invert static files root check

martenlienen:
Fixes #3108. I screwed up the check when I fixed the style issue.


mihran113 (approved):
Hey @martenlienen! Thanks for pushing the fix, I almost did it myself already.

Looks definitely good this time 😄

PR #3116, opened 2024-03-12
[feat] Allow HF callback to initialize Aim run at init end

dushyantbehl:
Current implementation of HF callback in Aim initializes the `Run` object at training begin time. This means if the users want to track any additional metrics they need to do so before training finishes while using `HFTrainer` object. 

This small change which initializes the `Run` object on `init_end` makes the run object available to user of this callback via `callback.experiment` and allows them to track any extra metrics or parameters with the same run other than the standard metrics output by HF.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Enhancements:
 - Replace grpc with http/ws as transport for aim tracking server (mihran113)
 - Remove `aim storage upgrade 2to3` command (mihran113)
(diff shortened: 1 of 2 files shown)


alberttorosyan (commented):
Hey @dushyantbehl. Thanks for your contribution 🙌
Changes look good!
Would you mind updating the `CHANGELOG.md` file according to the changes?

dushyantbehl:
@alberttorosyan sure!

dushyantbehl:
@alberttorosyan done with the changes, let me know if I need to change anything. 🥂

alberttorosyan (approved):

PR #3130, opened 2024-04-09
[fix] Resolve issue with multiple token refreshes

mihran113:


roubkar (approved):

aim/web/ui/src/services/api/api.ts
```diff
@@ -346,13 +348,20 @@ async function checkCredentials<T>(
...
+      window.location.assign(`${window.location.origin}/sign-in`);
       return parseResponse<T>(response);
     }
-    // Refresh token
-    const token = await refreshToken().call();
-    if (token) {
-      setAuthToken(token);
-      return refetch();
+    if (localStorage.getItem('refreshing') !== 'true') {
+      localStorage.setItem('refreshing', 'true');
```
alberttorosyan, line 355: Isn't there a possibility for race conditions here? Like, reading the 'refreshing' key right before it is set to 'true'?

alberttorosyan (approved):

PR #3141, opened 2024-04-26
Fix demo links

SandersAaronD:
Links to demos from readme were broken, this replaces those links with links taken from aimstack.io.


SGevorg (approved):
lgtm

Issue #2803, opened 2023-06-01
Option to hide/remove experiment names and tags when they have no associated runs


Nuno-Mota:
## 🚀 Feature

The feature would consist of two options, one for experiment names and one for run tags, that would allow hiding names/tags that have no associated runs (because runs were deleted, for example).

Potentially, there should also be an option to remove them entirely (this exists for tags, but not for experiment names, I believe).

### Motivation

The current behaviour can be confusing due to irrelevant experiment names / run tags, and also lead to cluttering of the UI, thus reducing its usability.

roubkar:
Hey @Nuno-Mota, sorry for the late response.
Thanks for this feature request.
Currently, we are focused on developing [Aim v4](https://github.com/aimhubio/aim#-aim-40).
After the initial release of v4 we're going to revisit also run groups / tags module and its functionality.

PR #3158, opened 2024-06-06
[feat] Add feature to delete full experiments

### LINKED ISSUE #2528: Delete experiments from UI
### Motivation

Runs can be deleted in a user-friendly manner from the UI. It would be helpful to do this for full experiments in order to not select a bulk of runs manually.

### Pitch

A delete option here:

image

mauricekraus:
This pull request introduces the ability to delete entire experiments directly from the user interface.

#### Delete Full Experiments:

Users can now delete entire experiments without having to manually select and delete individual runs.
This resolves the issue of having empty experiments after all runs are deleted, providing a more streamlined workflow.
Delete all runs of an experiment:

#### UI Consistency:

image

The delete functionality for experiments is currently available only on the experiment settings tab, making it consistent with the existing run detail  deletion process.

How we should delete experiments from the dashboard needs further discussion (due to the space constraints)

Issues Resolved:
Closes #2528 , #2803

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Enhancements:
+- Add feature to delete full experiments (mauricekraus)
+
 ## 3.20.1 Jun 3, 2024
 
 ### Enhancements:
--- a/aim/sdk/repo.py
+++ b/aim/sdk/repo.py
@@ -587,6 +587,25 @@ class Repo:
         else:
             return True, []
 
+    def delete_experiment(self, exp_id: str) -> bool:
+        """Delete Experiment data from aim repository
+
+        This action removes experiment data permanently and cannot be reverted.
+        If you want to archive experiment but keep it's data use `repo.get_experiment(exp_id).archived = True`.
+
+        Args:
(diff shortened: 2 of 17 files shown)


mauricekraus:
Sure, thanks for the heads up, i totally missed the remote part as well as the rockdb stuff.

mihran113:
Handling the remote part is unnecessary in this case  cause  repo will be local for the UI anyways. (It's for handling the run deletions from CLI and SDK)

mauricekraus:
@mihran113 this should resolve it, however, I'm still not sure about the remote thingy, at least locally this implementation gets rid of all runs in an experiment.

mihran113:
Hey @mauricekraus! Thank a lot for the changes. This should do the job. Regarding the remote: don't worry about it, I'll add that in the future when there'll be a need.  
I'll run some tests to see if everything works fine and merge the PR. In the meanwhile can I ask you to change the destination branch to the `aihmhubio:main` and update the changelog like this: 
```
## 3.21.0 

### Enhancements:
- Add feature to delete full experiments (mauricekraus)
```

mihran113 (approved):
Great job! Thanks a lot for the contribution. I've run some tests, everything looks fine! 🎉

Issue #3123, opened 2024-03-18
Aim UI can't open within Jupyterlab notebook


matrix72c:
## 🐛 Bug

In Jupyter Notebook, I can't open AimUI.

### To reproduce

After `%aim init` and `%load_ext aim`, `%aim up` comes to a error:

```
Launching Aim ...
Error: No such option: --force-init
```

### Environment

- Aim 3.19.1
- Python 3.10.8
- pip 22.3.1
- Ubuntu 22.04
- ipykernel 6.28.0
- jupyterlab 4.0.10

### Additional context

I find this problem related with https://github.com/aimhubio/aim/issues/2672 and hasn't been resolved for a while.

LmYjQ:
same issue
<img width="407" alt="image" src="https://github.com/aimhubio/aim/assets/12581433/9be1bd22-bce7-47f8-a48e-2a6ea4f37452">

aim version v3.19.3

synapticarbors:
I'm not familiar with the internals of `aim`, but in https://github.com/aimhubio/aim/blob/main/aim/cli/manager/manager.py#L54, the addition of the `--force-init` flag seems to be causing an issue. That flag no longer exists and looks like it was removed in https://github.com/aimhubio/aim/commit/3373cbbea1f9d213c3113fdaea2d2b16951dfe2b back in February.

PR #3162, opened 2024-06-13
[fix] Remove no longer available --force-init flag from cli manager

synapticarbors:
Currently the jupyter-lab extension is broken due to the `--force-init` flag that is added to the `aim up` call. See https://github.com/aimhubio/aim/issues/3123

This flag was removed from the cli in https://github.com/aimhubio/aim/commit/3373cbbea1f9d213c3113fdaea2d2b16951dfe2b

I've tested this with a local install of 3.20.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixes:
 - Increase websockets max_size for large images sent to server (jasonmads)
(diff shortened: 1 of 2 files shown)


alberttorosyan (approved):
@synapticarbors, changes look good! Thanks for your contribution.
I've updated the `CHANGELOG.md` file, proceeding with merge.

Issue #3177, opened 2024-07-01
--no Option for Aim Init, Do not Clear Existing Repo if it Exists


stevenjlm:
Happy to write a PR for this if we agree it helps.

## 🚀 Feature

An "-n" or "--no" option for `aim init`, so that it only creates a new repo if one does not already exist.

### Motivation

While running Aim in Kubernetes, I need to run aim init --repo /aim to make sure there is a repo for aim; however, since I have persistent storage, I would rather the --init option leave the repo intact if it already exists. Since this is happening in a container, I cannot answer "no" to the prompt.

### Pitch

There is already a "-y" or "--yes" option, I'm thinking of just adding "-n" or "--no" option.

### Alternatives

1. init's default could be set to initialize the repo only if it is not present.
2. There could be an entirely different command "soft-init" or maybe a option "--soft" to achieve the same effect.
I'm sure there are other alternatives that I have not thought of..

alberttorosyan:
@stevenjlm fair point.
w.r.t. naming of the new option; I think -s/--skip-if-exists describes the purpose of the flag better.

alberttorosyan:
On a separate note, @stevenjlm, would you be interested in contributing the fix to this issue?

stevenjlm:
Yes, I'm happy to do so. I'll work on a PR soon.

PR #3179, opened 2024-07-01
[feat] Add skip option in init commands

stevenjlm:
PR for #3177 
Adds a skip option for init in case the repo already exists.

Note: It does not seem like click has a standard way of dealing with conflicting options, so I decided to throw an error if "yes" and "skip-if-exists" are both specified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.22.1 Jul 4, 2024
+
+### Enhancements:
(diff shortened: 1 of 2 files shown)


alberttorosyan (approved):
@stevenjlm thanks for your contribution! LGTM
Could you please update the `CHANGELOG.md` too? I'll merge this PR right after that.

stevenjlm:
@alberttorosyan updated `CHANGELOG.md`

alberttorosyan (approved):

PR #3245, opened 2024-11-06
[fix] Resolve issues with unit test failures

mihran113:
Restrict `deeplake` upper version, because they had major changes in the new release and there's no documentation available yet for the new major version: 4.0.0


alberttorosyan (approved):
LGTM

Issue #3266, opened 2024-12-09
Run artifact_uri should return None if not set


sbatchelder:
## 🚀 Feature

If I do `run.artifact_uri` should return `None` if `run.set_artifact_uri(...)` has not been set yet. 

### Motivation

I'd like to check whether a run's  `artifact_uri` has been set yet or not, so as to know programatically whether it needs to be set or whether it's safe to run `run.log_artifact(...)`

### Pitch

In `run.artifact_uri` property function, after [base_uri](https://github.com/aimhubio/aim/blob/6feb6e7d0c7d1d90196d0f82ffcf029c20a9f491/aim/sdk/run.py#L422), add `if base_uri is None: return None`

### Alternatives

I could use the `try: ... except: ...` pattern whenever I go to log and artifact or check that artifact_uri is set, but that's a bit cumbersome.

mihran113:
Hey @sbatchelder! Thanks for opening the issue. The pitch makes a total sense. 
Would you like to make the changes yourself? It should be pretty easy to do so.

sbatchelder:
Happily :)
I'm also thinking of adding 
* a warning message if you set the URI without a scheme (just checks for `://`)
* ValueError in log_artifact if artifact_URI is `None`

PR #3268, opened 2024-12-13
[feat] Return `None` for `run.artifact_uri` if not set

sbatchelder:
#3266 
let run.artifact_uri return None if not set. 
Also raises useful warning and ValueError if uri improperly configured


mihran113 (approved):
Hey @sbatchelder! Thanks a lot for the PR. The changes look good. 
The checks are failing, but that's on our end :)

Issue #3259, opened 2024-12-02
PytorchLightning AimLogger custom contexts


sbatchelder:
## 🚀 Feature

The PytorchLightning AimLogger should allow users to define custom logged metric label mappings to aim contexts.
As it stands, `aim.sdk.adapters.pytorch_lightning.AimLogger` can only save metrics to Aim with the "subset" context (for train, val, test groups). The use of customizable prefixes is neat, but the fact that it's only to the "subset" context, that it's only for prefixes, and that the context mapping is defined in three separate init args is limiting. The logger should allow for any number of contexts and label-to-context-value mappings. 

### Motivation

I wanted to log the weighted, macro, and micro metric averaging methods using Aim's context mechanism. I'm using pytorch-lighting and I want to keep the code as lightning-friendly as possible. For instance, without a way to log `val_f1_weighted` to both the `subset:val` and `averaging:weighted` I'm left stuck with multiple aim-logged f1 metrics which should be grouped with contexts, but aren't.

### Alternatives

It's possible to invoke the `trainer.logger.experiment.track(..., context=....)` directly from wherever in the training/validation/test loops/callbacks in a lightning module so that you can log things to Aim with exacting specificity... but I shouldn't have to. It makes the code I write in the lightning modules less portable because it relies on implementation specific calls instead of the preferable built-in lightning logger mechanism.

### Pitch

Update or subclass AimLogger such that it can accept generic metric-label context mappings. Ideally without breaking backwards compatibility and retaining default functionality. 

#### Proposed Usage
```
logger = AimLogger(
    args.repo,
    context_postfixes = dict(
        averaging = {'macro': '_macro',       # for f1, precision, recall
                     'micro': '_micro', 
                     'weighted': '_weighted', 
                     'none': '_perclass'},  
        normalized = {'no': '_summed',        # for confusion matrix
                      'yes': '_normalized'},  
  ))

# then later in the lightning module callback functions
    def validation_step(self, ...):
        ...
        self.log("val_loss", loss)

    def on_validation_epoch_end(self):
        ...
        self.log('val_f1_macro', self.metrics['f1_macro'].compute(), on_epoch=True)
        self.log('val_f1_weighted', self.metrics['f1_weighted'].compute(), on_epoch=True)
```
Note how `self.log(val_loss, ...)` still works because the proposed AimLogger retains the default subset-context defaults, and in fact `self.log(val_f1_weighted, ...)` also benefits from the `val_` subset context mapping.

[...]

PR #3269, opened 2024-12-13
[feat] PytorchLightning AimLogger custom contexts

sbatchelder:
addresses #3259 "PytorchLightning AimLogger custom contexts"

There are two commits. First commit e0dc418 simply adds an AimLoggerWithContext subclass that uses the context configuring mechanism described in the associated Issue. Args `train_metric_prefix`, `val_metric_prefix`, `test_metric_prefix` ("SUBSET_metric_prefixes")  are removed.

The Second commit 15c9da5 is more extensive. It DOESN'T come with AimLoggerWithContext, instead refactoring the regular Lightning AimLogger to use the new context configuring mechanism. Care was taken make sure the changes are totally backwards compatible. SUBSET_metric_prefixes are present and functional but will raise a deprecation warning if input value is not default. 
I also went ahead and updated the documentation for the new schema. 

If the latter is too big a change, feel free to only integrate the first commit.

### DIFF
--- a/aim/pytorch_lightning.py
+++ b/aim/pytorch_lightning.py
@@ -1,3 +1,2 @@
 # Alias to SDK PyTorch Lightning interface
 from aim.sdk.adapters.pytorch_lightning import AimLogger  # noqa F401
-from aim.sdk.adapters.pytorch_lightning import AimLoggerWithContext  # noqa F401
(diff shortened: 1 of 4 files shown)


mihran113:
Hey @sbatchelder! Thanks a lot for the careful work you've done and for giving us the both options. I guess I should have weighted in under the issue, so you wouldn't have to make 2 separate solutions for the same purpose. Sorry for that. 
The second version looks a much better way to handle the problem at hand, and having the deprecation warnings in place will do the job to handle any previous and future usages. 
The changes look good at first sight, I'll just double check it myself and get back to you if there are any further comments.

mihran113 (approved):

PR #3274, opened 2024-12-22
[docs] pages/bookmarks intended section links

guspan-tanadi:


mihran113 (approved):
Hey @guspan-tanadi! Thanks a lot for opening the PR. The fix looks good.

PR #3282, opened 2025-01-18
chore: bump ruff version from 0.3.3 to 0.9.2 and fix some invalid/dead noqas

bluenote10:
I noticed that the ruff version was a bit outdated, and that there where quite a lot of spurious/dead `# noqa` lingering around.

This PR bumps ruff to the latest version, which has lead to a minor re-formatting of some code (mainly around fusing string literals).

I've also enabled a few ruff rules that have helped me to identify dead or malformed (`# noqa <code>` instead of the needed `# noqa: <code>` patter) noqa rules.

Regarding `F401`: Here we should stick to the library interface conventions from the [official specs](https://github.com/python/typing/blob/ad4cbc45f868f47031494694bdb7349119302f59/docs/spec/distributing.rst#import-conventions). Otherwise **type checkers will not understand the symbols exported in these modules**, which is a frequent paint point on user side, when libraries suppress `F401`. There are basically two options:

- Maintaining an `__all__` list manually.
- Using the `X as X` pattern to mark these symbols as explicitly public (and to differentiate them from actual unused imports).

I felt that the latter fits better to how Aim maintains its library interface.

My plan is to fully fix all `F401` in a follow-up, because the broken library interface is biting us.


aim/web/api/tags/views.py
```diff
@@ -41,7 +41,7 @@ async def search_tags_by_name_api(q: Optional[str] = '', factory=Depends(object_
             'id': tag.uuid,
             'name': tag.name,
             'color': tag.color,
-            'description' 'run_count': len(tag.runs),
```
bluenote10, line 44: This looks very much like an existing bug, no?

bluenote10:
Would it be possible to move forward with this? Is there anything I can do to help speeding things up / simplifying the review?

aim/web/api/tags/views.py
mihran113: This is very much an existing bug, thanks for pointing it out. 
I've pushed the fix for this in a separate PR, so can I ask you please to re-base with and remove this change?

mihran113 (approved):
Hey @bluenote10! 
Thanks for opening the PR and sorry for delayed response. Everything looks great, I've just left a comment with regards to the conflicts with main branch. I'll run the PR checks after the branch is up to date with main.

bluenote10:
> Everything looks great, I've just left a comment with regards to the conflicts with main branch. I'll run the PR checks after the branch is up to date with main.

@mihran113 Perfect, I've quickly resolved the tiny conflict, so this should be mergable again.

mihran113:
Thanks a lot for the changes @bluenote10! I'm going to merge the PR, there's a failure in unit-tests, but that's not related to your changes, I'm going to fix it separately.

Issue #2672, opened 2023-04-22
Aim UI does not open at all within Jupyterlab notebook with `%loadext`


adivekar-utexas:
## 🐛 Bug

When trying the [documented steps](https://aimstack.readthedocs.io/en/latest/using/jupyter_notebook_ui.html) to open the AIM UI on jupyter, it fails with an error.

### To reproduce

Install aim (I have `aim==3.17.3`)
```pip install -U aim```

From JupyterLab:

```
%load_ext aim

%aim up 
```

Passing "--repo=" also does not work:
```
%load_ext aim

%aim up   --repo="/Users/adivekar/my_project/"
```

In both cases, I get the following error. 

Notes:
- I am successfully able to open the repo by running
- The repo already has multiple runs in it
- This issue has been around since at least aim 3.8, I've tried multiple times over the last few months and never gotten it to work.

```
 /Users/adivekar/miniconda3/envs/eps2.2/lib/python3.10/site-packages/IPython/core/interactiveshel │
│ l.py:2369 in run_line_magic                                                                      │
│                                                                                                  │
│   2366 │   │   │   if getattr(fn, "needs_local_scope", False):                                   │
│   2367 │   │   │   │   kwargs['local_ns'] = self.get_local_scope(stack_depth)                    │
│   2368 │   │   │   with self.builtin_trap:                                                       │
│ ❱ 2369 │   │   │   │   result = fn(*args, **kwargs)                                              │
│   2370 │   │   │                                                                                 │
│   2371 │   │   │   # The code below prevents the output from being displayed                     │
│   2372 │   │   │   # when using magics with decodator @output_can_be_silenced                    │
│                                                                                                  │
│ /Users/adivekar/miniconda3/envs/eps2.2/lib/python3.10/site-packages/aim/ext/notebook/notebook.py │
│ :219 in execute_magic_aim                                                                        │
│                                                                                                  │
│   216 │   │   return                                                                             │
│   217 │                                                                                          │
│   218 │   # call corresponding handler                                                           │
│ ❱ 219 │   handlers[command](options, context)                                                    │
│   220                                                                                            │
│   221                                                                                            │
│   222 def load_ipython_extension(ipython):                                                       │
│                                                                                                  │
│ /Users/adivekar/miniconda3/envs/eps2.2/lib/python3.10/site-packages/aim/ext/notebook/notebook.py │
│ :168 in up                                                                                       │
│                                                                                                  │
│   165 │   │   │   display_id=True,                                                               │
[...]

PR #3319, opened 2025-04-06
[fix] Aim web ui integration in jupyter/colab

larissapoghosyan:
The PR #2696 added more robust mechanism of initializing aim server, checking the health, and only then enabling the web ui. However, there's no endpoint /projects/status/ now as part of api, so currently aim web ui integration for jupyter always fails.

This PR:
- Relies on status code of /projects endpoint instead
- Removes unnecessary delay in case of failure (for loop always calling sleep even if no tries left)
- Implements delayed retries with exponential backoff, and lets it to wait longer, as some google colab machines are slow and take at least 10-15 seconds to start the aim server.

Possibly addresses #2672 and #3123

Tested on both local jupyter notebooks, and google colab

### DIFF
--- a/aim/cli/manager/manager.py
+++ b/aim/cli/manager/manager.py
@@ -34,17 +34,17 @@ def run_up(args):
 
         server_path = 'http://{}:{}{}'.format(args['--host'], args['--port'], args['--base-path'])
         status_api = f'{server_path}/api/projects/'
-        retry_count = 5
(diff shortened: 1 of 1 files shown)


alberttorosyan (approved):
@larissapoghosyan, thanks for your contribution!
Approving this PR since the failure of code style checks is not related to this change.

Issue #3390, opened 2026-01-29
Aim crashing during initialisation of runs, cant view any runs anymore


ADH-LukeBollam:
## 🐛 Bug
I've recently started logging some figures each epoch, and am getting intermittent errors where I need to restart Aim or refresh the page until I can get it to load. Error messages in the browser say "Error: Response JSON already consumed" and I've attached one of the full logs below. I'll update this with the other issues as I hit them, there has been a few different types.

Today the latest error is preventing me from viewing any runs anymore, it seems to be hitting an exception trying to load run information and everything is green and in progress.

It's crashing at the line
```
    def list_corrupted_runs(self) -> List[str]:
        from aim.storage.encoding import decode_path

        def get_run_hash_from_prefix(prefix: bytes):
            return decode_path(prefix)

        container = RocksUnionContainer(os.path.join(self.path, 'meta'), read_only=True)

        return list(map(get_run_hash_from_prefix, container.corrupted_dbs))
```

when
container is <aim.storage.union.RocksUnionContainer object at 0x730ab670d080>
the prefix is b''
and paths is returning []

### To reproduce
Add some Plotly charts to logging each epoch, here's how I'm adding mine:
aim_fig = Figure(fig)
self.aim_logger.track(aim_fig, name=f'sample {i}', step=epoch)

### Environment

- Aim Version 3.29.1
- Python version 3.9.21
- pip version
- OS (e.g., Linux) Ubuntu
- Any other relevant information

### Additional context

Log:
IndexError: list index out of range
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/..../miniconda3/envs/v03/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/..../miniconda3/envs/v03/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
  File "/home/..../miniconda3/envs/v03/lib/python3.9/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/home/..../miniconda3/envs/v03/lib/python3.9/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
[...]

PR #3400, opened 2026-04-01
Fix: prevent IndexError in list_corrupted_runs for empty prefix

Regg819:
## Summary
Fixes a crash in `list_corrupted_runs()` when processing corrupted databases with empty prefixes.

## Bug
- **Issue**: #3390
- **Error**: `IndexError: list index out of range`
- **Cause**: `decode_path(prefix)` returns empty list `[]` when prefix is `b""`, and accessing `[-1]` on empty list crashes

## Fix
- Add check for empty path in `get_run_hash_from_prefix`
- Return `None` for empty paths and filter them out
- Prevents crash when corrupted DB has empty prefix entries

## Testing
This fix handles the edge case where corrupted_dbs contains entries with empty byte string prefixes.

<a href='https://ko-fi.com/J3J61X38VA' target='_blank'>Buy Me a Coffee at ko-fi.com</a>

### DIFF
--- a/aim/sdk/repo.py
+++ b/aim/sdk/repo.py
@@ -397,10 +397,13 @@ class Repo:
         from aim.storage.encoding import decode_path
 
         def get_run_hash_from_prefix(prefix: bytes):
-            return decode_path(prefix)[-1]
+            path = decode_path(prefix)
+            if not path:
+                return None
+            return path[-1]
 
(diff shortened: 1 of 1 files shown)


Regg819:
Hi team! Just checking in on this PR. Happy to address any feedback or make adjustments. Thanks!

